PR #102, opened 2019-09-16
Refactor a few things

stropitek:


stropitek:
Hello, I made a few changes
- Use immer to simplify state management and make sure no side effects are done on the state.
- Join 2 contexts together, so that one context contains all the state related to the displayed chart.
- Remove the `_` from state variables and pass them directly down the context instead of creating a new object.
- Refactor Crosshair so that's it renders in a separate layer and improves speed drastically (on my computer I went from 2-3 fps to 55fps! Same should be done for brush tool but it's not as easy. Trick is to the change the display only by changing the `transform` styling attribute. @hamed-musallam I think it would be valuable that you read [this article](https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas)
- Removed commented code

hamed-musallam:
@stropitek 
thank you very much for your help, but i have note, the Crosshair  is work fine as you mentioed when i open any example,but if you take Coffee as example, i open it and it's work perfect for the first time but when i zoom in  and then full zoom out the performance become worst.

stropitek:
Yes I noticed that I'm not sure why but I'm expecting this to be fixed once we have a better implementation of the brush tool

hamed-musallam:
> Yes I noticed that I'm not sure why but I'm expecting this to be fixed once we have a better implementation of the brush tool

but it's not related to brush tool, if you just double click by the mouse you will trigger full zoom out event and you will get the same result.

PR #129, opened 2019-10-31
very first step to include line broadening

djeanner:
I still have to include the real flesh of the line broadening, but I'm currently blocked by the fact that I don't see how to apply lilne broadening to the demo spectra.


hamed-musallam:
@djeanner 
i merge you changes in the master branch.

PR #133, opened 2019-11-03
Show and filter meta data from jcamp

michaelwenk:
Here added I a new panel "InformationPanel" which shows the parameters and values of a dummy object for now. Filtering the parameter list is also possible. The dummy object should be replaced by the meta data of a Datum1D object but I was not able to figure out how to access such meta data, i.e. for an active spectrum (which actually should be the desired one?).

### DIFF
--- a/src/component/panels/InformationPanel.jsx
+++ b/src/component/panels/InformationPanel.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect, useContext } from 'react';
 
 import {
   Table,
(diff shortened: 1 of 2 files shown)


src/component/panels/InformationPanel.jsx
```diff
@@ -7,25 +7,95 @@ import {
...
+  for (let content of tableContent) {
+    // check the property string only
+    let labelContent = content.children[0].innerHTML;
+    if (labelContent.match(regexp)) {
+      content.style.display = 'flex';
+    } else {
+      content.style.display = 'none';
+    }
+  }
+};
```
hamed-musallam: @michaelwenk
can you explain to my your code.
and also you should use Callback instead of regular function to prevent from initiate function object every time you call it.

you can call it what ever you want, and inside the function use setSearchSequence.


also at some case we have more than one spectrum and your code only handle one spectrum info data, so if we have more than one how you will handle it?

lpatiny: Indeed it is a problem when we have many spectra and I think spectrum info should be desactivate with a message if there is more than one spectrum selected. At least for now

hamed-musallam: -You not have to access to dom elements directly andthnig you need to update it to dom use state for that,
- you don't need to use regular expression here there are a short way for this using includes function in your case and it will give you the same result.

can you pealse check my last commit at you code and if you have question you can ask me when ever you want.

PR #136, opened 2019-11-05
line broadening / first implementation

djeanner:
To be tested. I had to define the test data using
        `re: new Float64Array([1, 1, 1, 1]),`
instead of
       `re: [1, 1, 1, 1],`
Not very elegant: test of returned values with toStrictEqual and values with 15 digits.
in lineBroadening.test.js
This is preliminary. We will need to check that indeed one Hz broadening is produced when using this function with a parameter of 1!!


lpatiny:
@djeanner  new Float64Array([1, 1, 1, 1]) is completely ok for testing and you should really make simple tests like that.
In case you want also (after making simple tests) make a complex test you can use 'toMatchSnapshot'. T-he first time it will create a snapshot of the result and then it will check that the result does not change. T-his is more a regression test.

djeanner:
fixed failing tests

PR #172, opened 2019-11-13
Add manual range analysis

michaelwenk:
Here, the RangesTablePanel was added. Like in case of integrals it is in the same table format.
I also added a subfolder "placeholder" and a file NoTableData.jsx. This was created to have a default view for empty tables and to use such class instead of copying and pasting the same source code every time. It is the same element which was already used, e.g. in integration table panel, if the table was empty.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3318,28 +3318,28 @@
           "dependencies": {
             "abbrev": {
               "version": "1.1.1",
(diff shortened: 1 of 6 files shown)


michaelwenk:
in addition: the panel was just created, so without any functionality for now

PR #182, opened 2019-11-20
use spectra-data to make auto peak picking Fix #170

### LINKED ISSUE #170: autopeak picking should use Spectra library
The button works but the result is not ok.

The auto peak picking should use the Spectra library and check which parameters we need to get a better result.

jobo322:


jobo322:
The test case for it is failing, if there is nothing to change for this method, I can refactor the test case

PR #195, opened 2019-12-03
First commit with master table and basic values

michaelwenk:
Here I created a master table with the basic information (from, to, integral). The rest has to be inserted next.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "react-table": "^7.0.0-beta.15",
     "react-use": "^13.9.0",
     "spectra-data": "^3.4.11",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -30,6 +30,7 @@
     "react-table": "^7.0.0-beta.15",
     "react-use": "^13.9.0",
     "spectra-data": "^3.4.11",
+    "styled-components": "4.4.1",
```
targos: we already depend on the `emotion` library for styling, please use it. It has a similar API to create styled components: https://emotion.sh/docs/styled

PR #196, opened 2019-12-05
Extended Ranges Table

michaelwenk:
Here, I implemented a ranges table with expandable signals as sub-table which also is expandable if there are coupling in there. This is actually a first draft of how display something from ranges in sub-tables. To expand a row one has to click on the toggle (black filled triangle) on the left side.  The deletion of ranges via the table is now possible too.


michaelwenk:
I also created a subfolder in "elements" to separate ReactTable (also the expandable one) from the others. A common style file was added too which is now used in both ReactTables.

PR #226, opened 2019-12-12
Highlight in ranges table and spectrum

michaelwenk:
Hey, here I implemented mainly a _highlight property in ranges which is used in Ranges.jsx and through the ReactTable components. If one hovers over or leave a ReactTableRow component then a handler is activated to do a given action (dispatch type). In case of ranges, the new added _highlight property of a range is set to true and that triggers to the corresponding range in the spectrum. Similar it is done for the ranges in spectrum. So each way works for now to highlight a range in the other component (table vs. spectrum).
The problem for now is that if we trigger/dispatch in that way then all components (at least in the ReactTables) are re-rendered everytime we hover or unhover a component. So the current state like expanded rows is reset and in origin (not expanded) form. 
Do you have any advises how to solve that or make it better/efficiently?


targos:
Adding an ephemeral property to the data doesn't seem ideal. Would it be possible to implement it with a dedicated state instead?
I'm thinking of a list of currently highlighted things (could be for example objects with `{ type: 'range', id: 'id_of_the_highlighted_range' }`) that is updated through actions. The places that need to update the highlights send actions and the places that need to show the highlights read the state.

michaelwenk:
> Adding an ephemeral property to the data doesn't seem ideal. Would it be possible to implement it with a dedicated state instead?
> I'm thinking of a list of currently highlighted things (could be for example objects with `{ type: 'range', id: 'id_of_the_highlighted_range' }`) that is updated through actions. The places that need to update the highlights send actions and the places that need to show the highlights read the state.

@targos Thanks for that help! 
I did such implementation and it works now as expected, I think. The highlight information is now stored as a state in the global context. For each existing spectrum object can now exist an array with objects to highlight. Each of them with a certain type and id, as you proposed.  
The ranges in spectrum as well as the ranges (rows) in the table are now sending and receiving the highlight information. The useMemo() is used for receiving/checking the highlight status. 

The ReactTableRow component is now open for all types of objects to highlight. The underlying row information just needs to contain the type and id information for an object.

michaelwenk:
I have to apologise in advance: there are a lot of duplicated commits because of several merges.

PR #233, opened 2019-12-19
Highlights between spectrum and tables for ranges, peaks and integrals

michaelwenk:
Here is an implementation for highlighting ranges, peaks and integrals between components in spectrum and ReactTable rows. For this, a separated and independent state "highlights" was introduced to be able to listen for and to push changes for components to highlight. 

Could one of you please also check the performance, i.e. for highlighting ranges?


targos:
We created a simple generic highlighting hook. See https://github.com/cheminfo/nmr-displayer/commit/e19766eb9eef1619221a814dfc4d9c7af8dc02d2

Here's how it should be used.

From within a component that should emit and/or listen to highlights, add a line like the following:

```js
const highlightIds = ['abc', 'def'];
const highlight = useHighlight(highlightIds);
```
`highlightIds` must be an array of strings or numbers. It will contain for example the unique range id.


`useHighlight` returns a object like so:
```ts
interface Highlight {
  isActive: boolean;
  show: Function;
  hide: Function;
  onHover: object;
}
```

- `isActive` will be `true` if any of the values in the `highlightIds` array is currently "shown".
- If the `show` function is called, all `highlightIds` transition to the "shown" state.
- If the `hide function is called, all `highlightIds` transition to the "hidden" state.
- `onHover` is a shortcut that can be passed to a component and is similar to `onMouseEnter={show} onMouseLeave={hide}`.

Example:

```js
function Tr(props) {
  const { value } = props;

  const highlight = useHighlight(value.id);

  return (
    <tr
      key={value.id}
      style={{
        backgroundColor: highlight.isActive ? 'red' : 'transparent',
      }}
      {...highlight.onHover}
    >
      <td>{value.id}</td>
    </tr>
  );
}
```

Live demo: https://cheminfo.github.io/nmr-displayer/#/test/highlight

michaelwenk:
I modified the code to use the new highlighting hook. The modifications from my side start at commit a11cfa6 ( https://github.com/cheminfo/nmr-displayer/pull/233/commits/a11cfa6dd65de89a9182f0cb237aa4b4b4e36a60 ).

lpatiny:
@hamed-musallam Please could you check and merge this pull request.

src/component/Notations/PeakNotation.jsx
```diff
@@ -198,55 +192,23 @@ export const PeakNotation = ({
...
-          handleOnEnterNotation(xIndex);
-          onMouseEnterAndLeaveHandler(ADD_HIGHLIGHT);
-        }}
-        onMouseLeave={() => {
-          handleOnMouseLeaveNotation();
-          onMouseEnterAndLeaveHandler(DELETE_HIGHLIGHT);
-        }}
+        onMouseEnter={() => handleOnEnterNotation(xIndex)}
+        onMouseLeave={() => handleOnMouseLeaveNotation()}
+        {...highlight.onHover}
```
targos: this will override the callbacks set above. if you need custom handling, you have to do something like this:

```js
onMouseEnter={() => {
  handleOnEnterNotation(xIndex);
  highlight.show();
}
onMouseLeave={
  () => handleOnMouseLeaveNotation();
  highlight.hide();
}
```

src/component/Range.jsx
```diff
@@ -0,0 +1,93 @@
...
+  const dispatch = useDispatch();
+
+  const deleteRange = useCallback(() => {
+    dispatch({ type: DELETE_RANGE, rangeID: id });
+  }, [dispatch, id]);
+
+  const DeleteButton = () => {
+    return (
+      <Fragment>
+        {`<!-- export-remove -->`}
```
targos: do we still need this?

src/component/elements/ReactTable/Elements/ReactTableRow.js
```diff
@@ -1,60 +1,17 @@
...
-        highlighted.find(
-          (highlight) =>
-            highlight.type === type && highlight.id === row.original.id,
-        )
-          ? HighlightedRowStyle
-          : null
-      }
-      {...row.getRowProps()}
+      css={highlight.isActive ? HighlightedRowStyle : null}
+      {...{ ...row.getRowProps(), ...highlight.onHover }}
```
targos:
```js
{...row.getRowProps()}
{...highlight.onHover}
```

targos:
there are conflicts to solve. could you clean up the PR and keep only one commit (no merge) ?

PR #287, opened 2020-01-14
fix ranges detection

jobo322:
reduced minMaxRatio option to 0.05


jobo322:
options like minMaxRatio should be accessible for users

PR #392, opened 2020-04-21
chore: try to read all data in bruker zip file

jobo322:


src/data/data1d/Data1DManager.js
```diff
@@ -44,23 +43,23 @@ export class Data1DManager {
...
     return datum1D;
   };
 }
 
 function getData(spectra) {
   let x =
-    spectra[0] && spectra[0].data && spectra[0].data && spectra[0].data.x
-      ? spectra[0].data.x
+    spectra[0] && spectra[0].data && spectra[0].data[0].x
+      ? spectra[0].data[0].x
```
lpatiny, line 54: This change would break the current version of jcamp converter. In which case to we have more than one element in the data array ? If it never happens the Bruker converter should be changed

src/data/Analysis.js
```diff
@@ -31,32 +31,75 @@ export class Analysis {
...
+        '2rr': BINARY,
+      };
 
-      // console.log(this.spectra);
-      // eslint-disable-next-line no-await-in-loop
-      files = await Promise.all(
-        files.map(async (file) => {
+      const zip = await jsZip.loadAsync(zipFile.binary);
+
+      let folders = zip.filter(function(relativePath) {
```
lpatiny, line 53: Please use arrow function

src/data/Analysis.js
```diff
@@ -31,32 +31,75 @@ export class Analysis {
...
+        if (relativePath.match('__MACOSX')) return false;
```
lpatiny, line 54: This line seems useless

src/data/Analysis.js
lpatiny: seems to me that what you want to do is simply:
```suggestion
     let folders = zip.filter( (path) => path.match( /(ser|fid|1r|2rr)$/ ));
```

src/data/data1d/Data1DManager.js
jobo322: So, I will upgrade jcampconverter dependency in brukerconverter

src/data/Analysis.js
jobo322: It is useful when the data is coming from macOS, some wrong things happen without it (in my short experience)

PR #418, opened 2020-05-01
feat: read processed 2D Bruker spectra from a zip file

jobo322:


src/data/data2d/Data2DManager.js
```diff
@@ -24,4 +24,24 @@ export class Data2DManager {
...
+
+    const datum2D = new Datum2D({
+      ...options,
+      meta: result.info,
+      data,
+      source: {
+        jcamp: null,
+        jcampURL:
+          options.source && options.source.jcampURL
+            ? options.source.jcampURL
```
lpatiny, line 39: How there can be a jcampURL if it is coming from Bruker ? I think this field should be null.

src/data/data1d/Data1DManager.js
```diff
@@ -63,7 +63,11 @@ function getData(spectra) {
     spectra[1] && spectra[1].data && spectra[1].data.y
       ? spectra[1].data.y
       : null;
-  // 2 cases. We have real and imaginary part of only real
 
-  return im ? XReIm.sortX({ x, re, im }) : XY.sortX({ x, re });
+  if (x[0] > x[1]) {
+    x.reverse();
```
lpatiny, line 68: Indeed this must be much much faster. Nice change !

src/data/data1d/Data1DManager.js
```diff
@@ -8,7 +8,7 @@ export class Data1DManager {
   static fromBruker = function fromBruker(result, options = {}) {
     const { info } = options;
     let data = getData(result.spectra);
-    if (data.im) info.isComplex = true;
+    if (!!data.im) info.isComplex = true;
```
lpatiny, line 11: In which case this is useful ?

jobo322: Currently, I am not sure which is a better solution to set isComplex in data came from brukerConverter. Or do you mean the !!?

lpatiny: Yes the '!!' . What is the value of data.im ? Why it was not working without '!!'

jobo322: you are just right

src/data/data2d/Data2DManager.js
jobo322: It is done already

PR #463, opened 2020-05-07
feat(phaseCorrection): add automatic phase correction in filter1D

jobo322:
I just did add the filter for auto phase correction. So we need to discuss about if the method just return phased data, ph0 and ph1 values.


lpatiny:
@hamed-musallam For may it may be merged but currently it will crash because it is based on an unreleased version of ml-spectra-processing.

PR #476, opened 2020-05-09
fix(autoPhase)

jobo322:
change noise level depending on the observed nucleus and reduce minRegSize parameters.

### DIFF
--- a/src/data/data1d/filter1d/autoPhaseCorrection.js
+++ b/src/data/data1d/filter1d/autoPhaseCorrection.js
@@ -11,13 +11,16 @@ export const name = 'automatic phase correction';
  * @param {number} [options.ph1=0]
  */
 
-export function apply(datum1D, options = {}) {
+export function apply(datum1D) {
   if (!isApplicable(datum1D)) {
     throw new Error('phaseCorrection not applicable on this data');
   }
-  let { minRegSize = 256 } = options;
+  let { nucleus } = datum1D.info;
 
-  return ReIm.autoPhaseCorrection(datum1D.data, { minRegSize });
+  let factorStd = nucleus === '1H' ? 3 : 4.5;
+  let options = { minRegSize: 30, factorStd };
+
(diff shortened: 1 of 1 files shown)


jobo322:
@lpatiny, this change needs a new release of spectra-processing

src/data/data1d/filter1d/autoPhaseCorrection.js
```diff
@@ -11,13 +11,16 @@ export const name = 'automatic phase correction';
...
-export function apply(datum1D, options = {}) {
+export function apply(datum1D) {
   if (!isApplicable(datum1D)) {
     throw new Error('phaseCorrection not applicable on this data');
   }
-  let { minRegSize = 256 } = options;
+  let { nucleus } = datum1D.info;
 
-  return ReIm.autoPhaseCorrection(datum1D.data, { minRegSize });
+  let factorStd = nucleus === '1H' ? 3 : 4.5;
```
lpatiny, line 20: The noise level should not be dependent of the nucleus but from the spectrum.
If could be a number of time STD or we have good results taking a number of time the MEDIAN value.

lpatiny: We have very diluted 1H sometimes (natural product extract) or highly concentrated 13C (quantitative carbon). We could evaluate here the noise level and give it to the auto phase correction algorithm.

jwist: I think we need to implement a function to determine the noise. The possibility to determine S/N is a useful feature. I suggest we can give a try to Damien's idea (https://doi.org/10.1002/mrc.4882). Maybe @djeanner can help.

It looks pretty simple to implement and give an idea of the quality of the spectra and an example can be found here thanks to Damien (https://github.com/Gr‐Jeannerat‐unige/san‐plot). It also has a nice graphical representation that can be displayed over the spectrum.

image

PR #503, opened 2020-05-20
support for JEOL NMR data

jwist:
allows support for JEOL NMR 1D data using nmr-parser package.


src/data/SpectraManager.js
```diff
@@ -33,6 +34,55 @@ export function addJcamp(spectra, jcamp, options = {}) {
...
 }
 
+export function addJDF(spectra, jdf, options = {}) {
+  // need to parse the jcamp
+  let converted = fromJEOL(jdf, {});
+  let info = converted.description;
+  let metadata = info.metadata;
+  delete info.metadata;
+  info.acquisitionMode = 0;
+  info.experiment = '1d';
```
lpatiny, line 44: This is expected to be the pulse sequence name of something like that. Not a fixed value because we don't know if it is 1d or 2d.

jwist: Yes, this is a debug error. It has been corrected and is now set according to data dimensions. I don't believe it is possible to make it according to pulse sequence name, since these can be edited by users (on the machines) and because most sequence can be run in 1D or 2D.

PR #528, opened 2020-06-24
Replace grey rectangle in multiplicity tree & enabling of assignment mode from multiplicity tree

michaelwenk:
- replacement of grey rectangle around multiplicity tree (as highlighting indicator) by controlled opacity and stroke width values
- enabling of assignment mode by clicking on a multiplicity tree
- fixed a problem with tree width and height
- simplifications


lpatiny:
Thanks !

PR #652, opened 2020-10-28
Introduction of correlation table

michaelwenk:
- addition of first version of correlation table to summary panel
- shows correlations between loaded 1D and 2D spectra and spectra types
- protons are attached to heavy atoms directly via HSQC or HMQC
- detection of simple problems only (for now)
- shows coloured atom labels if there is any problem was detected: 
- yellow: a proton is attached to more than one heavy atom, its label contains more than one label 
- blue: a proton is not attached
- red: the total atom count was reached and an atom is out of limit
- selection of molecular formula and setting of shift tolerances in same panel via buttons (for now)
- not editable at the moment
- not able to detect the number of attached protons for heavy atoms (information is missing at the moment)
- a simple overview to see whether atoms for an atom type are missing or error exist


src/component/panels/SummaryPanel/CorrelationTable.jsx
```diff
@@ -50,7 +51,7 @@ const getExperiment = (data) => {
   }
 
   pulse = pulse.toLowerCase();
-  if (pulse.includes('zg')) {
+  if (pulse.includes('zg') || pulse.includes('udeft')) {
     // + udeft ?
     return '1d';
   }
```
hamed-musallam: I have an idea to remove all the if statement and reduce the code

I think you can use regular expression directly inside of includes, the idea is to have a table containing return value as key and expression( regular expression represent every if statement) and loop throw it row by row, and if the condition is true return the key and break the loop... 

this change not mandatory but you have to move all the if statement to function and then call the function here

src/component/panels/MoleculePanel.jsx
```diff
@@ -196,7 +196,7 @@ const MoleculePanel = memo(
         if (displayerMode === '1D' && ranges && ranges.values) {
           return ranges.values;
         } else if (displayerMode === '2D' && zones && zones.values) {
-          return ranges.values;
+          return zones.values;
```
hamed-musallam: can you use DISPLAYER_MODE constant inside if letteral value 1d and 2d, you can import it from 
import { DISPLAYER_MODE } from './reducer/core/Constants';

src/component/panels/MoleculePanel.jsx
```diff
@@ -436,54 +434,15 @@ const MoleculePanel = memo(
 
     const handleOnUnlinkAll = useCallback(() => {
       data.forEach((datum) => {
-        // unlink in assignment hook state
         if (displayerMode === '1D') {
```
hamed-musallam: use DISPLAYER_MODE constant

src/component/panels/MoleculePanel.jsx
```diff
@@ -436,54 +434,15 @@ const MoleculePanel = memo(
...
-                axis: 'x',
-              },
-            }),
-          );
+          // unlink in assignment hook state
+          RangeUtilities.unlinkInAssignmentData(assignmentData, datum);
+          // unlink in global state
           const _datum = RangeUtilities.unlink(datum);
           dispatch({ type: CHANGE_RANGE_DATA, data: _datum });
         } else if (displayerMode === '2D') {
```
hamed-musallam: use DISPLAYER_MODE constant

hamed-musallam (changes requested):
First, I want to thank you for your contribution, can you check those changes request and once you finish let me know and then I will review it again

src/component/panels/SummaryPanel/CorrelationTable/Utilities.js
```diff
@@ -0,0 +1,275 @@
...
+      lodash
+        .get(error, 'ambiguousAttachment', [])
+        .some((index) => index === correlation.index)
+    ) {
+      return { color: 'orange' };
+    }
+  }
+
+  return undefined;
+};
```
hamed-musallam, line 85: I refactor this code for as an example of what you can do to write code better
{
const error =  lodash.get(state, `atomType.${correlation.atomTypes[0]}.error`, null) 
   if(error){
     const keys = [{key:"outOfLimit",color:"blue"},...........];
      for( let item of keys){
        if(Array.isArray(error[item .key]) && correlation.index in error[item .key]){
         return {color:item.color}; 
        }
       }
     return null;
    }

src/component/panels/SummaryPanel/CorrelationTable/Utilities.js
```diff
@@ -0,0 +1,275 @@
...
+    .filter((index, i, a) => a.indexOf(index) === i);
+  if (atomTypeToAttach === 'H') {
+    correlation.attached[atomTypeToAttach].forEach((indexToAttach, n) => {
+      if (lodash.get(correlations, `H[${indexToAttach}]`, false)) {
+        correlations.H[indexToAttach].label[`C${index + 1}`] = `H${index + 1}${
+          letters[n]
+        }`;
+      }
+    });
+  }
```
hamed-musallam, line 120: I prefer not to use lodash.get() function if have only one level of keys but it is a very useful tool if you have 2,3 or more keys path

hamed-musallam (changes requested):
you can take those notes as general notes for the whole code and for the  current and future changes

PR #731, opened 2020-12-08
feat: Slider component

hamed-musallam:


src/component/elements/Slider.jsx
```diff
@@ -0,0 +1,151 @@
...
+      }
+    `}
+  >
+    <FaAngleLeft
+      style={{ transform: `scaleX(${direction === 'right' ? '-1' : '1'})` }}
+    />
+  </div>
+);
+
+const transition = 0.45;
```
lpatiny, line 50: In the final result we don't see any animation which is ok. 
You should remove everything related to animations and transitions.
No need to use absolute positioning and CSS transforms.

src/component/elements/Slider.jsx
```diff
@@ -0,0 +1,151 @@
...
+        height: 100%;
+        width: 100%;
+        margin: 0 auto;
+        overflow: hidden;
+      `}
+      ref={ref}
+      key={width}
+    >
+      <div
+        className="sliderContent"
```
lpatiny, line 128: This class should be removed

lpatiny:
Please rename this component to 'NextPrev'. It was not obvious to understand what Slider was doing.

hamed-musallam:
> Please rename this component to 'NextPrev'. It was not obvious to understand what Slider was doing.
The slider is a common name used for this component

src/component/elements/Slider.jsx
hamed-musallam: I removed it but I add it because maybe in the future I need to apply styles using emotion js.

src/component/elements/Slider.jsx
hamed-musallam: the animation gives some beauty to the component, this is why I use it but the final decision, in the end, is up to you, if you need me to remove all the animation then I will remove it

lpatiny (commented):
Please rename the component to PrevNext
Arrows should disappear if first / last

PR #1007, opened 2021-04-30
feat(nmredata): refactor of importation/exportation

jobo322:
use the nmredata to import and export.


src/component/loader/DropZone.jsx
```diff
@@ -222,11 +218,24 @@ function DropZone(props) {
               },
             );
             break;
+          case FILES_TYPES.NMREDATA:
+            loadFiles(selectedFilesByExtensions, { asBuffer: true }).then(
+              async (files) => {
+                for (const zipFile of files) {
+                  const unzipResult = await Zip.loadAsync(zipFile.binary);
```
lpatiny, line 225: I would rather unzip the data in nmredataToNmrium function. The event is LOAD_NMREDATA_FILE so it is more logical to send directly the file which is in the case of a NMReData a ZIP.

lpatiny (commented):
Please add in util a folder named `nmredata` to put all the related files

lpatiny:
There is a conflict with your branch. Please merge the master in this branch and install Volta to be sure you are using the right version of `npm` (as I described in the Telegram chat)

PR #1040, opened 2021-05-10
chore: use usedColors variable per instance

hamed-musallam:


targos:
I haven't looked deeply at the code, but the approach seems fine.

hamed-musallam:
> I haven't looked deeply at the code, but the approach seems fine.

I define the used color at the level of the dispatch middleware

PR #1092, opened 2021-05-17
fix: always use the zones for polynomial baseline correction.

### LINKED ISSUE #1073: Polynomial baseline correction seems weird
Trying to select zones (shift + draw) for manual baseline correction

image

I obtain the following baseline correction

image

That seems weird. @jobo322 any idea where could be the problem.

jobo322:
close #1073


lpatiny (commented):
Did you check what happens when there is no zones selected (I guess it is ok) ?

How do you send the options for the polynomial degree ?  Seems there should be `options.regressionOptions`

jobo322:
> Did you check what happens when there is no zones selected (I guess it is ok) ?
> 
> How do you send the options for the polynomial degree ? Seems there should be `options.regressionOptions`

PR #1099, opened 2021-05-19
feat: improve default values for auto phase. close #1093

### LINKED ISSUE #1093: Automatic phase correction not working well
For this spectrum the automatic phase correction is not working well. I wonder what could be done to improve this.

[DP-1-1-SM.dx.zip](https://github.com/cheminfo/nmrium/files/6498708/DP-1-1-SM.dx.zip)

jobo322:


src/data/data1d/filter1d/autoPhaseCorrection.js
```diff
@@ -12,17 +12,17 @@ export const name = 'Automatic phase correction';
  */
 
 const defaultOptions = {
-  magnitudeMode: false,
-  minRegSize: 30,
-  maxDistanceToJoin: 256,
+  minRegSize: 10,
```
lpatiny, line 15: Could you change the variable name ? Is this minResionSize ?

src/data/data1d/filter1d/autoPhaseCorrection.js
```diff
@@ -12,17 +12,17 @@ export const name = 'Automatic phase correction';
...
+  maxDistanceToJoin: 128,
```
lpatiny, line 16: Should this option not be dependant from the number of points in the spectrum ? We have spectra with 1k points and some with 256k points.

src/data/data1d/filter1d/autoPhaseCorrection.js
```diff
@@ -12,17 +12,17 @@ export const name = 'Automatic phase correction';
...
+  magnitudeMode: true,
+  factorNoise: 5,
 };
 
 export function apply(datum1D, options = {}) {
   if (!isApplicable(datum1D)) {
     throw new Error('phaseCorrection not applicable on this data');
   }
   options = Object.assign({}, defaultOptions, options);
```
lpatiny, line 25: Please use the new syntax
`options = {...defaultOptions, ...options}`

lpatiny (changes requested):

src/data/data1d/filter1d/autoPhaseCorrection.js
jobo322: It should be tested because I believe it does not relate with the resolution of the spectrum, it is closer to how much noise has the spectrum, but I put a bigger cutoff for noise, I think we could test with those parameters to check if it works for most cases.

src/data/data1d/filter1d/autoPhaseCorrection.js
jobo322: I could

jobo322: So, we will need to wait for the new release of ml-spectra-processing

src/data/data1d/filter1d/autoPhaseCorrection.js
targos: you can also directly pass the spread to the function call: `reimAutoPhaseCorrection(datum1D.data, {...defaultOptions, ...options})`

PR #1106, opened 2021-05-21
Phase correction

lpatiny:
I tried to simplify manual phase correction in order to remove global variable and useRef.

@targos is this the correct way to use useRef ?

Also please could you help me so that we can still continue to make the phase correction even if we leave the button ?


lpatiny (commented):
This code still has a major problem : If you move the mouse continuously the spectrum will never be redrawn because the events only recalculate the phase that is relatively CPU intensive and the process to redraw apparently never has the hand to do his job.

Throttling is not really a solution because it would depend on the computer processing power.

@targos how to solve this in an elegant way ?

We should I think ignore the mouseMoveCallback if it was not yet redrawn. Kind of `if (notRedrawn) return`

src/component/elements/InputRange.jsx
```diff
@@ -24,135 +24,45 @@ const styles = {
...
-        step: stateValues.step === 'start' ? 'initial' : 'end',
-      };
-    }
-  };
+  function mouseUpCallback() {
+    window.removeEventListener('mousemove', mouseMoveCallback);
+    window.removeEventListener('mouseup', mouseUpCallback);
+  }
+
+  function mouseDownCallback() {
```
lpatiny, line 53: Should this be wrapped in a useEffect in order to be able to remove the event in all the cases ?

targos:
> This code still has a major problem : If you move the mouse continuously the spectrum will never be redrawn because the events only recalculate the phase that is relatively CPU intensive and the process to redraw apparently never has the hand to do his job.

I don't see any slowdown with this change and it seems to be very fast.

src/component/elements/InputRange.jsx
targos: do you think it's possible to not have the mouseup event?

targos:
There is a problem with the value in the PH0 and PH1 fields. It's only an integer now:
image

lpatiny:
> There is a problem with the value in the PH0 and PH1 fields. It's only an integer now:
> image

The integer is not really a problem. I changed the logic and if you want to have a finer control you can press shift to have fine tuning. But is seems that the degree level is enough most of the time.

lpatiny:
> > This code still has a major problem : If you move the mouse continuously the spectrum will never be redrawn because the events only recalculate the phase that is relatively CPU intensive and the process to redraw apparently never has the hand to do his job.
> 
> I don't see any slowdown with this change and it seems to be very fast.

It gives strange result on my computer at least because if you continuously move the mouse the spectrum never change. This is why in the old code there was a 50ms debouncing (or throttling, or just delay between 2 changes ?)

targos (dismissed):
need to discuss this so I understand the expected behavior

lpatiny:
At the functionality point of view it looks really good. @targos any obvious problem in the code ?

PR #1153, opened 2021-06-28
refactor-v5

Sebastien-Ahkrin:


src/component/elements/Input.tsx
```diff
@@ -0,0 +1,198 @@
...
+    textAlign: 'center',
+  },
+};
+
+export interface InputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
+  name?: string;
+  label?: string;
+  value?: any;
+  styleInput?: {
```
targos, line 28: You forgot to change the prop name in `ManualPhaseCorrectionPanel.jsx`, but I'm not sure why you renamed the prop to `styleInput`? Is it because of the collision with `InputHTMLAttributes` ? In that case you can do `InputProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'style'>`

src/component/elements/formik/FormikCheckBox.tsx
```diff
@@ -1,18 +1,27 @@
...
+  onChange?: (element: any) => void;
+  reverse?: boolean;
+}
+
+function FormikCheckBox(props: FormikCheckBoxProps) {
+  const {
+    label,
+    name,
+    onChange = () => null,
+    className = '',
```
targos, line 20:
```suggestion
    className = 'checkbox',
```

Maybe keep this default? I'm not sure if it's used.

src/component/elements/Input.tsx
```diff
@@ -48,7 +48,7 @@ const Input = forwardRefWithAs(
       label,
       value = '',
       name,
-      styleInput = { label: {}, input: {}, container: {} },
+      style: styleInput = { label: {}, input: {}, container: {} },
```
targos, line 51: can you also rename this?

src/component/elements/formik/FormikInput.tsx
```diff
@@ -24,7 +24,7 @@ function FormikInput(props: FormikInputProps) {
   const {
     label,
     name,
-    styleInput = { label: {}, input: {} },
+    style: styleInput = { label: {}, input: {} },
```
targos, line 27: and this?

targos:
There's a bug with some inputs. It's impossible to change their value.
For example ( with General -> 1H spectrum test) :
image
image

PR #1159, opened 2021-07-02
fix: simplify help provider to avoid re-renders

targos:


targos:
@hamed-musallam I open a pull request, because this change seems to break something completely unrelated on my computer:

image

Maybe an existing bug related to timing?

PR #1235, opened 2021-09-10
chore: added sample for ASAP HSQC of eserine

michaelwenk:


src/demo/samples.json
```diff
@@ -5,6 +5,10 @@
   {
     "groupName": "General",
     "children": [
+      {
```
targos, line 8: it probably shouldn't be the first element in the list

michaelwenk: @targos I moved it to the list's end

lpatiny:
We would need the original data (is it a jcamp ???). The parsing is done in the following project: https://github.com/cheminfo/nmr-parser

Please open a ticket in the nmr-praser project and upload the jcamp (as a ZIP) in the ticket. You may also try to make a PR with a testcase and fix the problem or ask @jobo322 for help.

PR #1239, opened 2021-09-14
refactor: update nmrium to use react-analysis-ui

Sebastien-Ahkrin:


Sebastien-Ahkrin:
* Error while you add a new molecule on the bottom right panel

lpatiny:
There is a pixel shift when you select a tool that changes the menu bar that should be fixed

targos:
- Fix the sidebar so it has the same size as the one on dev.nmrium.org. (in pixels, not percent)
- increase the size of the toolbar icons

src/component/NMRium.tsx
```diff
@@ -11,12 +12,10 @@ import {
   useRef,
   memo,
   Reducer,
-  CSSProperties,
   ReactElement,
   ReactNode,
 } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
-import SplitPane from 'react-split-pane';
```
targos, line 19: you can remove this dependency. it's not used anymore :)

src/component/NMRium.tsx
```diff
@@ -227,8 +219,8 @@ function NMRium({
       toggle(false);
     },
   });
-  const [isRightPanelHide, hideRightPanel] = useState(!initialShowPanels);
-  const [isResizeEventStart, setResizeEventStart] = useState(false);
+
+  const [isResizeEventStart] = useState(false);
```
targos, line 223: can we remove this state? it seems useless now if it cannot be changed

lpatiny: I don't remember in which case we are hiding on load the panel but it is something we could need in the future.

Not a must have before merging this PR

PR #1248, opened 2021-09-29
feat: update nmr-processing to 3.0.1

jobo322:
use the typescript version of nmr-processing


package-lock.json
```diff
@@ -7242,6 +7242,46 @@
...
+        "asynckit": "^0.4.0",
+        "combined-stream": "^1.0.8",
+        "mime-types": "^2.1.12"
+      },
+      "engines": {
+        "node": ">= 6"
+      }
+    },
+    "node_modules/jcampconverter/node_modules/nmr-processing": {
+      "version": "2.5.0",
```
targos, line 7259: we should update it in jcampconverter first to avoid the double dependency

src/component/panels/RangesPanel/RangesHeader.tsx
```diff
@@ -158,7 +158,7 @@ function RangesHeader({
       nucleus, // '19f'
       nbDecimalDelta, // 2
       nbDecimalJ, // 1
-      observedFrequency, //400
+      frequencyObserved, //400
```
targos, line 161: this rename seems like a mistake

PR #1268, opened 2021-10-15
feat: improvements correlation table, i.e. allow movement of links

michaelwenk:


hamed-musallam:
@michaelwenk 

These comments it's related to the UI and not to the core of your work,

1- validate the input so it's not accept empty value
image
2- remove Horizontal and vertical scroll
image
3-remove horizontal lines from   the cell seem that they appear only when hover
image
4- align and validate input fields and prevent accepting empty value
image
5- Prevent the table from expanding when we have fewer rows, as you can see the height of the rows become bigger when we have fewer rows, you have to take to not broke the scroll

PR #1319, opened 2021-12-16
fix: refactor nH parameter to integrationSum close #1318

### LINKED ISSUE #1318: Why nH ?
https://github.com/cheminfo/nmrium/blob/a46d63ceb7ad1008b07dac13b84c57246851fd4c/src/data/data1d/Spectrum1D/ranges/detectRanges.ts#L14-L15

I think this auto range picking also works with carbon and other elements.

It seems also that we need to know BEFORE calling the method the total number of elements.

@jobo322 Could you check why we call this nH and rename it to integration ? (better name ?)

@hamed-musallam Can we get the number of atoms before calling the peak picking ?

jobo322:


lpatiny (commented):
I guess this will need a migration script.
@hamed-musallam Could you tell how @jobo322 can write this ?

hamed-musallam:
We do not need to write code for migration because these options not
exported.

في الخميس، ١٦ ديسمبر ٢٠٢١ ١١:١٤ ص Luc Patiny ***@***.***> كتب:

hamed-musallam:
@lpatiny 
I'm going to merge this branch because as I mentioned in https://github.com/cheminfo/nmrium/pull/1319#issuecomment-995762097 that we do not need to write a migration script for these changes, is it ok for you? or do you still have notes?

PR #1339, opened 2022-01-24
Change the position of the molecule label and the size & scale of the molecule

Sebastien-Ahkrin:


Sebastien-Ahkrin:
@lpatiny i changed the position of the label, is this ok for you ? 

Capture d’écran 2022-01-24 à 15 36 09

Issue #1353, opened 2022-01-28
Warning with missing key


lpatiny:
When creating a range we get a warning in the console.

(take care I use the shortcut `r` to select the 'range picking' tool.

![2022-01-28 07 31 03](https://user-images.githubusercontent.com/1484241/151498886-1c49221a-6dd0-4d3c-8d78-d4529ba12812.gif)

react_devtools_backend.js:4061 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `RangesInner`. See https://reactjs.org/link/warning-keys for more information.
    at RangesInner (http://localhost:4000/src/component/1d/ranges/Ranges.tsx:24:3)
    at Ranges (http://localhost:4000/src/component/1d/ranges/Ranges.tsx:75:7)
    at svg
    at Chart1D (http://localhost:4000/src/component/1d/Chart1D.tsx:28:3)
    at div
    at MouseTracker (http://localhost:4000/src/component/EventsTrackers/MouseTracker.tsx:26:3)

PR #1355, opened 2022-01-28
Resolve missing key error

Sebastien-Ahkrin:
Closes: https://github.com/cheminfo/nmrium/issues/1353

### DIFF
--- a/src/component/1d/ranges/Ranges.tsx
+++ b/src/component/1d/ranges/Ranges.tsx
@@ -27,9 +27,8 @@ function RangesInner({
   return (
     <g clipPath={`url(#${displayerKey}clip-chart-1d)`}>
       {ranges?.values?.map((range) => (
-        <Fragment key={`container-${range.id}`}>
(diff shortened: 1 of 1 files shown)


src/component/1d/ranges/Ranges.tsx
```diff
@@ -27,7 +27,7 @@ function RangesInner({
   return (
     <g clipPath={`url(#${displayerKey}clip-chart-1d)`}>
       {ranges?.values?.map((range) => (
-        <>
+        <Fragment key={`container-${range.id}`}>
```
targos, line 30:
```suggestion
        <Fragment key={range.id}>
```

The "container-" doesn't make the key more specific in this case.
You can also remove the key from `<Range>` below

Sebastien-Ahkrin: i can't do that :/
if i remove the "container" prefix and the key in Range

The application will just enter in an infinite loop while i try to add a range with shift + click

Issue #1351, opened 2022-01-27
Resolve TODOs about d3-path


targos:
There are a few TODOs that are all about the same thing.

We should build the svg paths using d3 instead of manual string construction.

https://github.com/d3/d3-path/blob/main/README.md#path (no need to add a new dependency, the methods should be available in the `d3` package)

For example:
https://github.com/cheminfo/nmrium/blob/efe0f0a4b4f94df2535fc51390956f5320db840b/src/component/1d/Line.tsx#L46-L57

lpatiny:
@targos Just wondering if d3 is the best implementation (anyway this reduce is not ok neither). 

https://github.com/d3/d3-path/blob/main/src/path.js

It seems to me that d3 creates huge amount of intermediate path strings. I would rather use the old array join trick.

Please take care that the current code is rather tricky because you slice pathPoints.x but not point.y (and therefore you have this weird `y[i + 1]`

So I propose the following kind of code (didn't test it) that doesn't slice (waste of time / memory) and doesn't create intermediate strings.

```js
let path = [];
path.push(`M ${_scaleX(pathPoints.x[0])} ${_scaleY(pathPoints.y[0])} `);
for (let i=1; i<pathPoints.x.length; i++) {
  path.push(`L ${_scaleX(pathPoints.x[i])} ${_scaleY(pathPoints.y[i])} `);
}
return path.join(' ');
```

Reference: https://jsbench.me/fxkll021qz

targos:
@lpatiny I'm fine with not using d3-path, but I would like to have a similar internal API to simplify the code.
Something like:

```js
const pathBuilder = new PathBuilder(); // internally create the array
pathBuilder.moveTo(_scaleX(pathPoints.x[0]), _scaleY(pathPoints.y[0])); // internally push a M instruction to the array
for (let i=1; i<pathPoints.x.length; i++) {
  pathBuilder.lineTo(_scaleX(pathPoints.x[i]), _scaleY(pathPoints.y[i])); // internally push a L instruction to the array
}
return pathBuilder.toString(); // internally return array.join(' ')
```

What do you think?

lpatiny:
For me both are readable and should give the same performance so I have no preferences.

targos:
Reopening because of https://github.com/cheminfo/nmrium/pull/1358#pullrequestreview-868046029

PR #1360, opened 2022-02-01
refactor: change moveTo to use directly value instead of a variables

Sebastien-Ahkrin:
Closes: https://github.com/cheminfo/nmrium/issues/1351

### DIFF
--- a/src/component/2d/tools/slicing/VerticalSliceChart.tsx
+++ b/src/component/2d/tools/slicing/VerticalSliceChart.tsx
@@ -38,13 +38,14 @@ function VerticalSliceChart({
 
       const pathPoints = xyReduce({ x, y });
 
-      const lastXIndex = pathPoints.x.length - 1;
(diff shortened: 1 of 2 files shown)


src/component/2d/tools/slicing/VerticalSliceChart.tsx
```diff
@@ -38,12 +38,13 @@ function VerticalSliceChart({
...
-
       const pathBuilder = new PathBuilder();
 
-      pathBuilder.moveTo(scaleY(pathPoints.y[lastYIndex]), scaleX(lastXIndex));
+      pathBuilder.moveTo(
+        scaleY(pathPoints.y[pathPoints.y.length - 1]),
+        scaleX(pathPoints.x.length - 1),
+      );
+
       for (let i = pathPoints.x.length - 2; i > 0; i--) {
```
targos, line 48: >= 0 here too

Issue #1341, opened 2022-01-26
Part of the structure may be hidden under the title of next tab


lpatiny:
If in the general preferences you activate the Database tab

![image](https://user-images.githubusercontent.com/1484241/151119971-cc9643f9-81bb-48d6-8387-4caf8d6e4f5c.png)

For large molecule the bottom of the molecule is hidden under the title of the 'Database' tab

![2022-01-26 08 19 31](https://user-images.githubusercontent.com/1484241/151120043-4a058c4d-08fd-48ab-b9da-e4aed2e2e015.gif)

PR #1361, opened 2022-02-01
Change the dom to not allow the "Structures" panel to expand

Sebastien-Ahkrin:
Closes: https://github.com/cheminfo/nmrium/issues/1341

@lpatiny Actually i have that (with this fix)
Capture d’écran 2022-02-01 à 16 25 38

Do this works with your example ?

### DIFF
--- a/src/component/panels/MoleculesPanel/MoleculePanel.tsx
+++ b/src/component/panels/MoleculesPanel/MoleculePanel.tsx
@@ -144,6 +144,7 @@ function MoleculePanelInner({
                 </span>
                 <div
                   css={styles.slider}
+                  className="mol-svg-container"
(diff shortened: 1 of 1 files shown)


lpatiny:
Seems to me it is not fixed.

Just draw a long vertical molecule and put a `O` in the bottom.

Sebastien-Ahkrin:
Actually, i can get something like that:
Capture d’écran 2022-02-03 à 11 39 47

:)

targos:
Nice, now the size is correct by default, but there is still an issue when the container becomes smaller (for example if you open other panels).

image

Usually this is fixed by putting `min-height: 0` on the right flex elements.

targos:
See here. You need `min-height: 0` on two more elements and it works: 
image

PR #1376, opened 2022-02-15
feat: exclusion zones filter

hamed-musallam:


targos:
image

You are using the wrong version of node/npm

test-e2e/panels/structures.test.ts
```diff
@@ -12,6 +12,9 @@ test('should draw structure and display it with MF', async ({ page }) => {
 
   // Click on the "Add Molecule" button.
   await nmrium.page.click('data-test-id=panel-structures-button-add');
+
+  await nmrium.page.waitForSelector('#drawarea1');
```
targos, line 16: This is not correct. If the selector is not here yet when we try to click on it, the correct solution is to create a locator for it and use `locator.click()`

https://playwright.dev/docs/api/class-locator#locator-click

hamed-musallam: the problem happen because of the scale transition, for now I will wait for 500 ms until the transition is finished, do you have a better solution for this

hamed-musallam: one potential solution is to replace the scale animation with fade and this will work without the need to wait. Can we check whether the animation is finished?

PR #1392, opened 2022-03-01
Allow multiple hybridizations & added correlation data to NMRiumData interface & enable to edit 2D signals and set path lengths & new filter button in correlation table

michaelwenk:


src/component/reducer/actions/LoadActions.ts
```diff
@@ -46,10 +46,36 @@ function setData(
       tolerance: DefaultTolerance,
     });
   } else {
-    draft.correlations = correlations;
+    // in case of older NMRium data are imported, convert hybridization string to number array
+    // @TODO remove following command to overwrite correlations at some point in future
+    draft.correlations =
```
lpatiny, line 51: For me we don't need to support a migration. This was an experimental feature so it may be breaking.

lpatiny:
The functionality is ok. @targos does the code looks of to you ?

PR #1393, opened 2022-03-01
feat: save NMRium preferences by workspace

hamed-musallam:


hamed-musallam:
@lpatiny 
Can you test it

lpatiny:
The functionality is ok for me.

@targos Here is the context of this PR. Currently on www.nmrium.org the exercises show all the options of NMRium which is not ok. The reason for this is that preferences are stored in local storage and when we reload the preferences it ignores the one specific to exercises (ie the exercise 'mode').

However there is also a nice feature in NMRium that is the 'mode'. You can define a preset of settings that will configure NMRium in the mode you like.

With this PR the goal is that settings are related to the mode. In the future this would allow a user to create its own mode in the general preferences depending if he prefers to make an assignment of 2D or just a prediction.

This PR also allows to hide the 'general settings' so that in the exercise mode there is no way for the student to add more features.

Finally this mode approach will allow to have our NMRium version specific to prediction as we discussed.

@targos could you review the implementation ?

targos:
I see issues with the implementation (mostly that the state update functions are not pure, since they interact with localstorage), but that's not new from this PR, so it looks ok.

src/component/hooks/usePanelPreferences.ts
```diff
@@ -0,0 +1,11 @@
+import { useMemo } from 'react';
+
+import { usePreferences } from '../context/PreferencesContext';
+
+export function usePanelPreferences(key: string) {
+  const preferences = usePreferences();
+
+  return useMemo(() => {
+    return preferences.formatting.panels[key] || null;
+  }, [key, preferences.formatting.panels]);
```
targos, line 10: nothing takes time or risks creating a new object here, the useMemo is counterproductive (comparing the dependencies is not costless)

PR #1416, opened 2022-03-07
refactor: zoom

hamed-musallam:


lpatiny:
Horizontal zoom does not zoom the integral anymore.

hamed-musallam:
@lpatiny 
i should work right now

PR #1467, opened 2022-04-06
refactor: buttons

hamed-musallam:


src/component/header/HeaderActionButtons.tsx
```diff
@@ -0,0 +1,41 @@
...
+    applyLabel = 'Apply',
+    cancelLabel = 'Cancel',
+    applyDataTestId,
+  } = props;
+  return (
+    <div style={{ marginLeft: 10, marginRight: 10, display: 'flex' }}>
+      <Button.Done
+        onClick={onApply}
+        borderRadius="0.3rem"
+        data-test-id={applyDataTestId}
```
targos, line 26: We normally don't need a test id for these cases. You should be able to use `text=label` in the test's selector

hamed-musallam: @targos 
thank you, https://github.com/cheminfo/nmrium/pull/1467/commits/4f1df76c14a2ab9640059d834ad47a06ded867f7

hamed-musallam:
@lpatiny 
I replaced all the buttons and use the new button component, can you check it before I merge this branch.
@targos do you have any comments on this PR?

PR #1470, opened 2022-04-11
refactor: assignments tool

hamed-musallam:


hamed-musallam:
@michaelwenk 
can you check with me why when the mouse comes over the atom, it is not highlighted the ranges, I tried to trace your way but I still do not catch the full image

lpatiny:
The functionalities to assign signals and zones seems to work correctly.

Issue #1471, opened 2022-04-11
Ranges / Preview publication string is not working


tomlebl:
Preview publication string in the ranges panel is not working. It works fine in v0.23.0 
The code fails in function saveAsHTMLHandler in  RangesHeader.tsx because preferences.current.formatting.nucleusByKey is an empty object. 
Workaround: Open and save General Settings that corrects data stored in local storage nmr-general-settings and error disappears 

Likely, initial preferences state is not getting set properly

hamed-musallam:
Hello @tomlebl, it seems that there is a bug when updating the nmr-general-settings when moving from the old to the new version of nmr-general-settings in the local storage, Can you send us the old nmr-general-settings?

tomlebl:
@hamed-musallam I have just tested the current HEAD and all seem to be good.

PR #1472, opened 2022-04-11
fixing issue #1471

tomlebl:
I am not entirely sure if this is the best way of fixing it but the fix does clear the bug.


hamed-musallam:
@tomlebl 
we solved the bug from the root in 42a20a1e1b6f2f576a25e2e28a1c4e1fd6f789b0

PR #1525, opened 2022-05-03
feat: float molecule structures over the spectrum

hamed-musallam:


src/component/elements/draggble/SVGDRaggable.tsx
```diff
@@ -0,0 +1,82 @@
+import { useEffect, ReactFragment } from 'react';
```
lpatiny, line 1: Please fix the name of the file to SVGDraggable. Because it is just a case change please use `git mv SVGDRaggable.tsx SVGDraggable.tsx` otherwise it may be problematic

hamed-musallam: done

Issue #1545, opened 2022-05-13
Zones and preferences


lpatiny:
When defining zones we are dealing with 2 nucleus. One for the horizontal axis and one for the vertical axis.

In NMRium we have example spectra that are 1H,1H (cosy) and spectra that are 1H,13C (hsqc, hmbc) (we could have other nucleus as well).

In the preferences of the zones we need to deal with the nucleus separately. This means that we should display the list of all the independent nucleus and set the preferences for each nucleus (and not couple of them). In a cosy there should actually be only 1H.

![image](https://user-images.githubusercontent.com/1484241/168213075-6e4870fe-bb38-418e-911b-1cc8640592ab.png)

For an 1H,13C we should therefore see a first group for 1H and a second group for 13C.

hamed-musallam:
closed by 88eee3c2758b66c0556e52f8933e8f9fa6d5b849

Issue #1524, opened 2022-05-03
Not able to set the format of delta in peaks preferences


lpatiny:
In peak panels preferences it is not possible to set the format for delta in ppm

![image](https://user-images.githubusercontent.com/1484241/166455887-3d40081a-0adf-4528-bb4a-50c39e2c1c68.png)

hamed-musallam:
@lpatiny 

we hide it for a purpose, we format the value based on the nucleus from the general setting

https://github.com/cheminfo/nmrium/blob/c071f5ad9143be406a8cb19e743e2fb03444a40e/src/component/panels/PeaksPanel/PeaksPreferences.tsx#L168

https://github.com/cheminfo/nmrium/blob/c071f5ad9143be406a8cb19e743e2fb03444a40e/src/component/panels/PeaksPanel/PeaksTable.tsx#L32
https://github.com/cheminfo/nmrium/blob/c071f5ad9143be406a8cb19e743e2fb03444a40e/src/component/panels/PeaksPanel/PeaksTable.tsx#L81-L92

lpatiny:
Indeed this is an excellent point. Need to think about it if we should keep this feature because in Range it is not used for example.

<img width="551" alt="image" src="https://user-images.githubusercontent.com/1484241/166735875-7b80b712-24a3-4992-bd68-46bd20533525.png">

@hamed-musallam Could you check (without doing any change) what would be the impact of remove this general preference and keep it at the level of each panel ?

hamed-musallam:
We format the number based on Nucleus in different components

Footer for 1d and 2d
![image](https://user-images.githubusercontent.com/35760236/168158560-fa1db06e-538a-4901-8e82-8af538b78895.png)
![image](https://user-images.githubusercontent.com/35760236/168158743-8e0ba5f0-a5a1-4962-95f5-b644c0453813.png)

Multiple spectra analysis table column header
![image](https://user-images.githubusercontent.com/35760236/168159152-bda3b7a4-3da0-4b07-8613-5281e79a17aa.png)

Peaks table and annotations 
![image](https://user-images.githubusercontent.com/35760236/168159479-eabf7b80-07b3-416d-a1c6-145b1195b993.png)

Search input field in database panel 
![image](https://user-images.githubusercontent.com/35760236/168159720-6843e55e-3146-4e1c-b310-e5232a380821.png)

Range  edit modal (delta,coupling)
[...]

PR #1547, opened 2022-05-13
feat:  panel preferences

hamed-musallam:
Refactor panels preferences based on #1524 , #1545 and integral width and colour preferences by nucleus 
image


hamed-musallam:
@lpatiny 

can you check the functionality?

lpatiny:
There is no way to format the coupling constant in Hz in the range panel, the option is missing

image

lpatiny:
Please rename 

Nuclei formatting -> Number formatting for crosshair and info line.

image

lpatiny:
In Ranges, please format the 'delta (Hz)' even for a range using the 'Hz' formatting.

image

Please rename in the preferences or ranges:

image

From -> From (ppm)
To -> To (ppm)
Absolute -> Absolute integration
Relative -> relative integraton

lpatiny:
In zones seems the preferences change nothing at all.

image

In does not make sense to have Absolute and Relative for each nucleus. The integration is expected to be the integration under the full drawn zone.

So in the preferences of zones we should have:
- For each nucleus, 1H, 13C, etc the preferences for '∂ (ppm)' (just formatting, no way to hide)
- Globally (not for each nucleus): 'Absolute integration', 'Relative integration'

The possibility to display the columns to display the integrations.

hamed-musallam:
@lpatiny 

how to calculate the integration of the relative and the absolute value for zones?

lpatiny:
> @lpatiny
> 
> how to calculate the integration of the relative and the absolute value for zones?

For now just remove everything related to integral on Zone. We don't calculate it for now.

PR #1550, opened 2022-05-13
feat: allow edition of equivalence for hetero atoms

michaelwenk:


michaelwenk:
@lpatiny @hamed-musallam This PR is to use the new version of nmr-correlation package and to enable the edition of equivalences for hetero atoms in summary panel.

PR #1561, opened 2022-05-20
feat: preview spectrum from database 

hamed-musallam:


lpatiny:
Preview of the spectrum with mouse hover works perfectly. We could decide about the color and vertical position in the preferences of the database panel but this can be done later and @hamed-musallam you could create an issue for it.

When adding the spectrum (click +):

image

It is still however the recreated spectrum that is added even if the jcamp is available.

Also (and this should be fixed first) if you try to click on the '+' for the second spectrum there is a crash

image

image

hamed-musallam:
@lpatiny 

resurrect spectrum from ranges failed, 
image

hamed-musallam:
@lpatiny 

i finished this PR, i add color and margin-bottom preferences and I add the jcamp when clicking on the + button if the jcampURL is available.

 What remains, is the bug when we resurrect spectrum from ranges

Issue #1578, opened 2022-06-07
Database and clicking on spectra


lpatiny:
Hope this can be done without hacks otherwise we should reconsider this feature.

From the database panel when we click on a line the spectrum should be added and stay even we leave the row.

We could select many rows as well !!!! So we can click to select, click to unselected but we can apply this on many rows.

As vertical position we should use the default behaviour from the 'stack', see the related issue: 
 https://github.com/cheminfo/nmrium/issues/1577

It would be nice to also have different colors for the spectra.

lpatiny:
This is temporary.

Indeed I just saw you have margin bottom (px) and color already in preferences.

I don't know what is the best way to proceed and it depends what is the simplest to develop. You may as well keep this property (margin bottom) and add a 'space' property to have the first spectrum position and then the distance between spectra.
As for the color for now they could have all the same color.

hamed-musallam:
@lpatiny 

is it acceptable to you if we add the spectrum to the spectra list like when we press on the '+' button and remove it once the user clicks again on the row, and keep the current functionality for mouse over as it is currently?

lpatiny:
If it is easier to implement it looks like a good approach.

hamed-musallam:
resolved by 8e8b0f2

PR #1588, opened 2022-06-20
feat: pin/unpin spectrum when clicking on the database row

hamed-musallam:


lpatiny:
Functionalities are ok thanks !

It would be nice to make the difference between 'hover' and 'selected' spectra. I think the simplest would be to have to intensity (opacity) of the same color.

Strong color for selected, light color for hover.

Currently when you are over a spectrum and you click there is no visual difference to highlight the fact it is selected. After it would change the opacity.

hamed-musallam:
@lpatiny @targos 

actually, i am aware of this and I did it like that because i can not hover and change the background for all rows because of td with rowspan and even if I use the sibling selector it will not work well, i have nothing to do. This is basic HTML/CSS behaviour unless I solve the problem manually by adding and removing the class based on the onMouseEnter/onMouseLeave events.


i saw others use tbody element as a container for the group of rows and I do not know if it's correct to have multiple tbody.

https://codepen.io/paulofsis/pen/kVEVYw

lpatiny:
One way that could solve this issue is that when selecting a row you can add in the `<tr>` a 'data-selected' property

In this case you can use CSS do deal with 3 cases:
- :hover and data-selected not present
- :hover and data-selected is present
- data-selected is present (and not hover)

lpatiny:
If I try to load the solvant database I get

image

lpatiny:
@hamed-musallam are you able to finalize this branch.

Also why the same approach that you are using now (selecting many rows in the table) can not be used to pin a spectrum ?

hamed-musallam:
@lpatiny 

we have an old discussion about this branch what was missing at that time is how to deal with rows with a rowspan, At the end, we agreed on using <tbody> around the <tr>,  which was need refactoring on the react-table that we use in the whole project

hamed-musallam:
> If I try to load the solvant database I get
> 
> image

it seems the problem from the merge, it seems the conflicts is not solved in a proper way

hamed-musallam:
@lpatiny 

https://github.com/cheminfo/nmrium/issues/1578
https://github.com/cheminfo/nmrium/issues/1578#issuecomment-1156581271

Issue #1569, opened 2022-06-03
Generate image of the spectrum from outside component


hamed-musallam:
in addition to exporting spectra as SVG, jpeg we need a way to capture and save them directly to the external data service. This has to be done from where our component is imported. for example, in Chemotion ELN they need to be able to capture the Blob of the generated image and store it  as a blob in the database or store the image in the file server and store its path in the database

hamed-musallam:
@lpatiny 

The user can click somewhere outside NMRium or trigger an event in the  NMRium Wrapper to generate an image as a blob to save it somewhere else, maybe in a database or as upload it as a file in another file server.


I will ask Chandu to comment on this ticket

CS76:
Use case: ELNs and repositories would like to create a thumbnail or preview of the spectra for the user to quickly identify the spectra and not have to load the spectra every time. 

Trigger point:
Ideally, this will be triggered behind the scenes by the parent application (via the nmrium wrapper) and NMRium just has to pass back the image/png or image/jpg blob or SVG back as a response. This will then be handled by the parent application (saved to the file system on the server or to a database).

PR #1589, opened 2022-06-21
feat: generate blob for the spectra from an outside component

hamed-musallam:
@targos 
we use useImperativeHandle to customize the instance value that is exposed to parent components when using ref so it is possible to call the generate a Blob function, and then they can do whatever they need with this blob, to store it in the database or save as png or svg  ... etc in an external file server.

https://github.com/cheminfo/nmrium/issues/1569

### DIFF
--- a/src/component/NMRium.tsx
+++ b/src/component/NMRium.tsx
@@ -165,7 +165,7 @@ const defaultData: NMRiumData = {
 };
 
 export interface NMRiumRef {
-  generateBlob: () => BlobObject | null;
+  getSpectraViewerAsBlob: () => BlobObject | null;
 }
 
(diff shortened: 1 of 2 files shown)


targos:
Implementation LGTM but can we find a better name for the exposed method?

hamed-musallam:
@targos 

> Implementation LGTM but can we find a better name for the exposed method?

I really do not know, this is what comes to my mind, we can call it getView() or something similar

targos:
The name can contain the word "Blob" because it's part of what is returned, but I think it should also have a hint about what the blob contains (an image? of what?)

hamed-musallam:
@targos 

getSpectraViewerBlob or generateSpectraViewerBlocb ?

hamed-musallam:
@targos 
thanks, I will make the changes and then  merge the PR

PR #1602, opened 2022-06-30
feat: search in database by molecule structure

hamed-musallam:


src/data/data1d/database.ts
```diff
@@ -36,19 +38,32 @@ export interface InitiateDatabaseResult {
...
   const search = (keywords: string | string[] = []) =>
     filter(data, { keywords });
+  const searchByStructure = (molfile: string) =>
+    prepareSearchByStructure(moleculesDB, molfile);
 
-  return { data, getSolvents, search };
+  return { searchByStructure, data, getSolvents, search };
+}
+
+function prepareSearchByStructure(
```
lpatiny, line 60: Please rename to `processSearchByStructure^

src/data/data1d/database.ts
```diff
@@ -36,19 +38,32 @@ export interface InitiateDatabaseResult {
...
+  moleculesDB: MoleculesDB,
+  molFile: string,
+): DatabaseNMREntry[] {
+  const molecule = OCL.Molecule.fromMolfile(molFile);
+  const result = moleculesDB.search(molecule) || [];
+  return result.map((entry) => entry.data, []);
```
lpatiny, line 66: Please remove this empty array that is useless.

lpatiny (changes requested):
Preparing the database is relatively slow and we should not do it onload but only the first time we open the substructure search dialog.

src/data/data1d/database.ts
```diff
@@ -36,19 +38,32 @@ export interface InitiateDatabaseResult {
...
 
-  return { data, getSolvents, search };
+  return { searchByStructure, data, getSolvents, search };
+}
+
+function prepareSearchByStructure(
+  moleculesDB: MoleculesDB,
+  molFile: string,
+): DatabaseNMREntry[] {
+  const molecule = OCL.Molecule.fromMolfile(molFile);
```
lpatiny, line 64: In the callback of the editor we receive back a molecule and we should not change it to a molfile and then recreate a molecule. We are loosing information during this process !!!!

Issue #1604, opened 2022-07-01
use idCode in database lookup


lpatiny:
Currently molfile is used to store the query in the database lookup. This is not a good idea because we are loosing some information.

IDCode should be used and the new version of react-ocl allows to set and retrieve this property.

To set the value when starting the editor:

https://github.com/zakodium-oss/react-ocl/blob/main/src/components/StructureEditor.js#L10

To get back the idCode on change there is now a third parameter in the callback:

https://github.com/zakodium-oss/react-ocl/blob/main/src/components/StructureEditor.js#L77

@hamed-musallam could you change the code to use idCode ?

PR #1609, opened 2022-07-04
feat: use idCode in database lookup

hamed-musallam:
close: #1604


hamed-musallam:
@lpatiny 

I think we have a bug in the structure editor when I clear the structure I got this error.

 how to check if the id Code is empty or not?

image

lpatiny (commented):
I added a line to prevent idCode to be null and for me the functionnalities are ok and you may merge this branch.

PR #1611, opened 2022-07-05
feat: improve general settings

hamed-musallam:


lpatiny:
LGTM. @hamed-musallam just try to figure out why some tests are not passing.

hamed-musallam:
@lpatiny 

This bug will go once we merge it, but we have to discuss if we need to hide the tool completely from the general setting, do we have any case we need that?

lpatiny:
Most of the workspace have no tools at all

image

This seems strange to me.

hamed-musallam:
@lpatiny 

this is just because the local storage did not update, I have to change the local storage version number and it will be reset with the new changes.

you can remove the data from storage and reload the NMRium

hamed-musallam:
@lpatiny 

can you test the functionality?

lpatiny:
The functionalities work as expected.

I'm trying to figure out what the 'Reset' button is doing

image

Apparently it resets the settings for the current workspace.

I think it should be moved to the top next to the 'workspace' drop down menu and renamed to 'Reset workspace settings'. Clicking on this button should not close the dialog box.

The current `Reset` button should be renamed to `Cancel` and put back the original settings when the dialog was opened.

hamed-musallam:
@lpatiny 

to make the behavior consistent any change happens to be permanent you should click on the save button, in case of reset it is reset the properties visually (temporary in front of the user) and to make it permanent you have to click on the Save button.

Issue #1620, opened 2022-07-12
Live preview of baseline correction


lpatiny:
Guido was asking if we could see the result of the baseline correction before applying it ?

There could be a live preview when we change the parameters and we could 'Cancel' or 'Apply' it.

PR #1621, opened 2022-07-12
feat: live preview for baseline correction

hamed-musallam:
#1620


lpatiny:
I tried with polynomial correction and each time I add a zone I expect to see new spectrum but it does not work.

hamed-musallam:
@lpatiny this is another story, I made the live update just when change the options in the header. I will implement it also when you add zones

lpatiny:
Functionality is ok now. But why when we click on baseline correction we have vertical rescale. Quite annoying this and I would prefer it does not happen.

hamed-musallam:
@lpatiny 

can you test it again?

hamed-musallam:
@lpatiny 

can you test this PR again and check also the phase correction

lpatiny:
Why was this PR closed ?

There is still a little problem of zoom out in case you don't apply the changes.

hamed-musallam:
@lpatiny 

It is closed by mistake.

hamed-musallam:
@lpatiny 
Do you mean when we press cancel? is this in all the filters or just in the baseline correction?

PR #1623, opened 2022-07-13
feat: add lorentz to gaussian window function 

jobo322:
There is code for Lorentz to Gaussian window function to use as a filter like `linebroadening`. I think the panel of `zero filling` should change to add this functionality or add a new icon for anodization. Please let me know I need to change anything


src/data/data1d/filter1d/lorentzToGauss.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const id = 'lorentz-to-gauss';
+export const name = 'Lorentz-to-Gauss Window';
+
+// This window function is based on https://spin.niddk.nih.gov/NMRPipe/ref/nmrpipe/gm.html
+export function apply(datum1D: Datum1D, options: any = {}) {
+  if (!isApplicable(datum1D)) {
+    throw new Error('LG not applicable on this data');
+  }
+  const { center = 0, GB: gaussHz = 0, LB: expHz = 0 } = options;
```
lpatiny, line 11: Why do we need to rename the variables (moreover uppercase for a variable). I think they should be named everywhere gaussHz and expHz.

lpatiny (commented):
@hamed-musallam Could you help Alejandro to finalize this PR and add the parameters in the header bar

image


@jobo322 could you check that the order of applying the filters is correct and the zero filling is done at the end ?

src/data/data1d/filter1d/lorentzToGauss.ts
hamed-musallam: @lpatiny @jobo322 

can you let me know what are the new options that we need to add to the header?

jobo322: We need a drop menu with the name of the functions, At the moment we have only `exponential` and `lorentz-to-gauss` 

- exponential only need one value:

`expHz` number [-inf, inf]

- lorentz-to-gauss needs three parameters

`center` number in 0-1
`gaussHz` number in 0-1
`expHz` number [-inf, inf]

hamed-musallam:
@jobo322 @lpatiny 

I can see only one new filter lorentzToGauss but where is exponential?

do we have to create new panel options for it or we should include these options in the zero filing and line broadening options?

what is the order of this filter?

PR #1633, opened 2022-07-20
feat: 2d fid

hamed-musallam:
for testing purpose


lpatiny:
Could you check that this file do no crash ?

[16.zip](https://github.com/cheminfo/nmrium/files/9672578/16.zip)

It crash as soon as you move the mouse.

PR #1647, opened 2022-07-27
feat: apodization window preview

hamed-musallam:


lpatiny:
Toggling live preview does not work

image

hamed-musallam:
@targos 

can you check why the Netlify deploy no longer works

targos:
Ok, it's because of the branch rename. I changed the configuration in Netlify to watch the `main` branch but I think you need to push something to trigger a new build of this pull request.

PR #1683, opened 2022-08-08
Optimize peaks

lpatiny:


src/data/data1d/Spectrum1D/peaks/optimizePeaks.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function optimizePeaks(datum1D: Datum1D, options: OptimizePeaksOptions) {
+  const { from, to } = options;
+  let { re, x } = datum1D.data;
+
+  if (from !== undefined && to !== undefined) {
+    const fromIndex = xFindClosestIndex(datum1D.data.x, from);
+    const ToIndex = xFindClosestIndex(datum1D.data.x, to);
+
+    x = x.slice(fromIndex, ToIndex);
```
lpatiny, line 22: Why making a copy rather than subarray as suggested in the issue ?

src/data/data1d/Spectrum1D/peaks/optimizePeaks.ts
```diff
@@ -0,0 +1,29 @@
...
+interface OptimizePeaksOptions {
+  from: number;
+  to: number;
+}
+
+export function optimizePeaks(datum1D: Datum1D, options: OptimizePeaksOptions) {
+  const { from, to } = options;
+  let { re, x } = datum1D.data;
+
+  if (from !== undefined && to !== undefined) {
```
lpatiny, line 18: How could from or to be undefined if in the type definition they are mandatory and numbers ?

lpatiny:
Currently I didn't succeed to let it work.

We should anyway only allow optimization of maximum 8 peaks at the same time. This means that you should zoom if there are more than 8 peaks. An alert message should be displayed in this case.

src/data/data1d/Spectrum1D/peaks/optimizePeaks.ts
hamed-musallam: we need it in one case when we full zoom out, but in any way this case can not happen because I always pass the window range

src/data/data1d/Spectrum1D/peaks/optimizePeaks.ts
hamed-musallam: this is my mistake

hamed-musallam:
> Currently I didn't succeed to let it work.
> 
> We should anyway only allow optimization of maximum 8 peaks at the same time. This means that you should zoom if there are more than 8 peaks. An alert message should be displayed in this case.

it tested with 8 peaks and it is crash on my side

lpatiny:
We can limit to 4 peaks for now. I think for the applications we want it is good enough. We could have a similar function at the level of Range / Signal in the future.

lpatiny:
The current version yields to strange results. I guess it is related to the refactor of nmr-processing.



@jobo322 are you testing the new nmr-processing library locally ? There may be some issues in the way you are testing.

jobo322:
sorry in the last version of nmr-processing, I forgot to convert width to PPM for xyPeaksOptimization, in the other hand my function to convert width Hz/PPM was not pure. 
version 9.0.2 of nmr-processing will contains the fixing.

lpatiny:
@hamed-musallam please add a tooltip 'Optimize peaks' like for the other tools

lpatiny:
@hamed-musallam Do we sent the shape with the new values for the simulation of the shape ?
 


Changing FWHM has no impact and I'm wondering why. @jobo322 maybe you could advice.

hamed-musallam:
@lpatiny 

yes it should be, I always update the original data

Issue #1635, opened 2022-07-21
Import of JEOL raw data


ak-jeol:
Loading JEOL raw data files (1D and 2D) results in an error.

1D raw import error
The x and y arrays must have the same length
Error: The x and y arrays must have the same length

2D raw import
It does not load the data, but there is no error

1D processed
No problem

2D processed
Cannot read properties of undefined (reading 'experiment')
TypeError: Cannot read properties of undefined (reading 'experiment')

Attached is ZIP file with 1D, 2D, parameter format, functionalized process lists

[nmrium.zip](https://github.com/cheminfo/nmrium/files/9159472/nmrium.zip)

ak-jeol:
Jose -

The parameter x_points for the data set I sent you was 14993.

The complex array consists of real+imaginary data points that translates to a total of 14993*2 = 29986.

Regards

Ashok

jobo322:
@ak-jeol many thanks I will check why the converter returns 15000 instead of 14993

jobo322:
@ak-jeol, checking the converter I found that we are using dataPoints instead of the parameter 'x_points':

- should those values be the same number? 
- is it ok to read this value as `uint32` as we are doing [here](https://github.com/cheminfo/jeolconverter/blob/master/src/parseJEOL.js#L58)

if you have any documentation about it, I could go further to fix this. many thanks again

ak-jeol:
Jose –

You can find this parameter at byte offset 176 and you read in a 4 byte unsigned integer.

Please refer to the data format document I sent in the previous email.

Regards

Ashok

From: Ashok Krishnaswami
Sent: Friday, July 29, 2022 12:26 PM
To: cheminfo/nmrium ***@***.***>; cheminfo/nmrium ***@***.***>
Cc: Mention ***@***.***>
Subject: RE: [cheminfo/nmrium] Import of JEOL raw data (Issue #1635)

Jose –

Yes, x_points would be the parameter to use to extract the number of data points.

Basically, our dimensions are x,y,z,a,b,c,d,e

So, for 2 dimensional spectra you would look at x_points (similar to F2 in BRUKER lingo) and y_points (similar to F1) for the direct and indirect dimension.

In term of reading in values, the fact that you read in our processed 1D data set fine would indicate that unit32 would be ok.

The programmer went out to lunch, so I will ask when he gets back.  He can look at your source code and let me know.

I have attached our data format documents.

Regards

Ashok

jobo322:
[...]

PR #1685, opened 2022-08-09
fix: round data to dataPoints reported

jobo322:
close: #1635

### DIFF
--- a/src/data/data1d/Data1DManager.ts
+++ b/src/data/data1d/Data1DManager.ts
@@ -41,11 +41,11 @@ export function fromCSD(result, options = {}, usedColors = {}) {
   let offset = dimension.coordinatesOffset.magnitude;
 
   let buffer = dependentVariables[0].components[0];
-  let re: number[] = [];
-  let im: number[] = [];
-  for (let i = buffer.length - 1; i > 0; i -= 2) {
-    re.push(buffer[i - 1]);
-    im.push(buffer[i]);
+  const re = new Float64Array(n);
+  const im = new Float64Array(n);
+  for (let i = buffer.length - 1, index = 0; i > 0; i -= 2) {
+    re[index] = buffer[i - 1];
+    im[index++] = buffer[i];
   }
 
   let data: any = {};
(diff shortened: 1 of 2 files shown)


src/data/data1d/Data1DManager.ts
```diff
@@ -41,11 +41,11 @@ export function fromCSD(result, options = {}, usedColors = {}) {
   let offset = dimension.coordinatesOffset.magnitude;
 
   let buffer = dependentVariables[0].components[0];
-  let re: number[] = [];
-  let im: number[] = [];
-  for (let i = buffer.length - 1; i > 0; i -= 2) {
-    re.push(buffer[i - 1]);
-    im.push(buffer[i]);
+  const re = new Float64Array(n);
```
lpatiny, line 44: This code change nothing right ? Just optimization ?

Could you also change the x variable so that it is a Float64Array ?

Could you also avoid the .map for data.im. We have a utility to multiply a xArray

Could you add a comment to this function. What is CSD ? (should this method be renamed ?)

lpatiny: Ok I understood. It is the length of the array that is based on dimension.count and not on the buffer anymore.

targos: I feel like this code should be somewhere in the parser, not in nmrium

lpatiny: Yes I agree. @jobo322 why is this code not in nmr-parser ?

We have a lot of changes in the data structure in NMRium while the goal on nmr-parser is to avoid this seems to me.

https://github.com/cheminfo/nmrium/blob/94e9d90fdb1b0ddf90ff25b3755ce35bf7682682/src/data/SpectraManager.ts#L95-L111

In nmr-load-save do we still plan to use nmr-parser ?

jobo322: nmr-load-save will export the nmrium format

jobo322: > Ok I understood. It is the length of the array that is based on dimension.count and not on the buffer anymore.

@lpatiny yes, dimension.count could be bigger than buffer length (small difference), so the idea is to fill with zeros. 

Is strange that `count` and data length do not match, julien did use `count` and he was working with a guy that understands JEOL data.

Issue #1692, opened 2022-08-11
Processing playwright testcase


lpatiny:
Check that you can do:
- Apodization
- Zerofilling
- Fourier transform

https://user-images.githubusercontent.com/1484241/184158313-edf6e57e-4e75-441d-bdad-6059763f3287.mp4

lpatiny:
Thanks Hamed ! Indeed @wadjih-bencheikh18 Please update the existing test:

- to check that there is a scale with the right beginning and end value
- to check the number of elements in the 'Filters' panel
- to check if there is a spectrum that is displayed (just a big path)

PR #1695, opened 2022-08-12
Processing 13c spectrum test case

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/1692


test-e2e/panels/filters.test.ts
```diff
@@ -79,3 +84,32 @@ test('process 1d FID 13c spectrum', async ({ page }) => {
     await baselineCorrectionFilter(nmrium);
   });
 });
+
+test('Processing 13c spectrum', async ({ page }) => {
```
lpatiny, line 88: You don't need to create a new test but just append the new steps to the previous step.

test-e2e/panels/filters.test.ts
```diff
@@ -57,7 +57,12 @@ async function baselineCorrectionFilter(nmrium: NmriumPage) {
     ),
   ).toBeVisible();
 }
-
+async function xScaleDomain(nmrium: NmriumPage, min: number, max: number) {
```
lpatiny, line 60: Please add a method in NmriumPage called assertXScaleDomain(min, max).

We will often reuse it ! And don't hesiate to add new method in NmriumPage

test-e2e/panels/filters.test.ts
```diff
@@ -57,7 +57,12 @@ async function baselineCorrectionFilter(nmrium: NmriumPage) {
...
+  const firstTick = nmrium.page.locator('.x >> .tick >> nth=0');
```
lpatiny, line 61: You could try to refactor to use something like:
```js
const xTicks = page.locator(‘.x >> .tick’)
const firstTick = xTicks.first()
const lastTick = xTicks.last()
```

test-e2e/panels/filters.test.ts
wadjih-bencheikh18: but in the previous test there is more filter used , can i remove them?

lpatiny: You may leave them if there are more filters.

wadjih-bencheikh18: okey thank you

Issue #1693, opened 2022-08-11
Molecule test case improvement


lpatiny:
We have currently the following test:

https://github.com/cheminfo/nmrium/blob/f271adef416c34bac8cfdebde76617c9626d25b3/test-e2e/panels/structures.test.ts#L41-L42

And we need couple of improvements.

After drawing the first molecule check that the SVG is there
Add a second molecule
Check C6H12
check SVG is there
Go back on the first molecule
Go next to second molecule
Copy to text and check the clipboard contains the correct text file (take a snapshot).

https://user-images.githubusercontent.com/1484241/184193019-2aba09de-7ebe-4f46-9062-692b8ec9e600.mp4

PR #1696, opened 2022-08-12
Improve molecule test

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/1693#event-7173446797

### DIFF
--- a/test-e2e/panels/structures.test.ts
+++ b/test-e2e/panels/structures.test.ts
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
   // Open the "Structures" panel.
   await nmrium.clickPanel('Structures');
 
-  // The SVG container should not be rendered when there are no molecules.
-  await expect(nmrium.page.locator('.mol-svg-container ')).toBeHidden();
-  // Click on the "Add Molecule" button.
-  await nmrium.page.click('data-test-id=panel-structures-button-add >> nth=0');
-
-  // Select the "aromatic ring" tool.
-  await nmrium.page.click('canvas >> nth=0', {
-    position: {
-      x: 35,
-      y: 200,
-    },
+  await test.step('empty panel', async () => {
+    // The SVG container should not be rendered when there are no molecules.
+    await expect(nmrium.page.locator('.mol-svg-container ')).toBeHidden();
   });
-  // Draw the aromatic ring.
-  await nmrium.page.click('canvas >> nth=1', {
-    position: {
-      x: 50,
-      y: 50,
-    },
+  await test.step('Add aromatic ring molecule', async () => {
+    // Click on the "Add Molecule" button.
+    await nmrium.page.click(
+      'data-test-id=panel-structures-button-add >> nth=0',
+    );
+
+    // Select the "aromatic ring" tool.
+    await nmrium.page.click('canvas >> nth=0', {
+      position: {
+        x: 35,
+        y: 200,
+      },
+    });
+    // Draw the aromatic ring.
+    await nmrium.page.click('canvas >> nth=1', {
+      position: {
+        x: 50,
+        y: 50,
+      },
+    });
+    // Save the molecule.
+    await nmrium.page.click('text=Save');
   });
-  // Save the molecule.
-  await nmrium.page.click('text=Save');
 
-  // The molecule SVG rendering should now be visible in the panel.
-  await expect(
-    nmrium.page.locator('.mol-svg-container #molSVG0'),
-  ).toHaveAttribute('xmlns', 'http://www.w3.org/2000/svg');
+  await test.step('Check the visibly of the aromatic ring molecule', async () => {
+    // The molecule SVG rendering should now be visible in the panel.
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toHaveAttribute('xmlns', 'http://www.w3.org/2000/svg');
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toBeVisible();
 
-  // The molecular formula should now be visible in the panel.
-  await expect(nmrium.page.locator('text=C6H6 - 78.11 >> nth=0')).toBeVisible();
+    // The molecular formula should now be visible in the panel.
+    await expect(
+      nmrium.page.locator('text=C6H6 - 78.11 >> nth=0'),
+    ).toBeVisible();
 
-  // The number of molecules should now be visible in the panel.
-  await expect(nmrium.page.locator('text=1 / 1 >> nth=0')).toBeVisible();
+    // The number of molecules should now be visible in the panel.
+    await expect(nmrium.page.locator('text=1 / 1 >> nth=0')).toBeVisible();
+  });
+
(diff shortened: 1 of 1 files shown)


test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+
+    // Check the visibility.
+
+    // The molecule SVG rendering should now be visible in the panel.
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toHaveAttribute('xmlns', 'http://www.w3.org/2000/svg');
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toBeVisible();
```
targos, line 87: This is not testing the second molecule. I guess it should be `#molSVG1`?

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
-  // Draw the aromatic ring.
-  await nmrium.page.click('canvas >> nth=1', {
-    position: {
-      x: 50,
-      y: 50,
-    },
+  await test.step('Add aromatic ring molecule', async () => {
+    // Click on the "Add Molecule" button.
+    await nmrium.page.click(
+      'data-test-id=panel-structures-button-add >> nth=0',
```
targos, line 17:
```suggestion
      'data-test-id=panel-structures-button-add',
```

Avoid using `nth=0` if the locator is guaranteed to resolve to only one element.
It's also good to remove the `nth` selector so that if because of a bug there are more elements, the selector will throw an error.

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toHaveAttribute('xmlns', 'http://www.w3.org/2000/svg');
+    await expect(
+      nmrium.page.locator('.mol-svg-container #molSVG0'),
+    ).toBeVisible();
+
+    // The molecular formula should now be visible in the panel.
+    await expect(
+      nmrium.page.locator('text=C6H6 - 78.11 >> nth=0'),
```
targos, line 49:
```suggestion
      nmrium.page.locator('text=C6H6 - 78.11'),
```

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+    ).toBeVisible();
+
+    // The number of molecules should now be visible in the panel.
+    await expect(nmrium.page.locator('text=1 / 1 >> nth=0')).toBeVisible();
```
targos, line 53:
```suggestion
    await expect(nmrium.page.locator('text=1 / 1')).toBeVisible();
```

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
-  await nmrium.page.click('text=Save');
 
-  // The molecule SVG rendering should now be visible in the panel.
-  await expect(
-    nmrium.page.locator('.mol-svg-container #molSVG0'),
-  ).toHaveAttribute('xmlns', 'http://www.w3.org/2000/svg');
+  await test.step('Add a second molecule and check the visibility', async () => {
+    // Click on the "Add Molecule" button.
+    await nmrium.page.click(
+      'data-test-id=panel-structures-button-add >> nth=0',
```
targos, line 59:
```suggestion
      'data-test-id=panel-structures-button-add',
```

Please remove all other `nth=0` that are not necessary.

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+    );
+
+    // Select the "aromatic ring" tool.
```
targos, line 62:
```suggestion
    // Select the "ring" tool.
```

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+    await nmrium.page.click('canvas >> nth=0', {
+      position: {
+        x: 35,
+        y: 180,
+      },
+    });
+    // Draw the aromatic ring.
```
targos, line 69:
```suggestion
    // Draw the ring.
```

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,119 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
+  });
+  await test.step('Switch between molecules', async () => {
+    // Start with the second molecule.
+    await expect(nmrium.page.locator('text=2 / 2 >> nth=0')).toBeVisible();
+    await expect(
+      nmrium.page.locator('text=C6H12 - 84.16 >> nth=0'),
+    ).toBeVisible();
+
+    // Go to the previous molecule.
+    await nmrium.page.click('.react-shadow-8g1a46-Arrow');
```
targos, line 105: You cannot use `react-shadow-xxx` classes as selectors because they are not stable and can change anytime.
In this case, I think you can use a React selector. Something like this: `_react=Arrow[direction="left"]`

targos (commented):
It's good job, but there is one wrong check where you verify the first molecule instead of the second one.
To make sure that a test is working as expected, I recommend you to run it in "debug" mode one last time before doing the commit. In this mode, you can go step by step in the test and check visually that everything works and that the assertions are done on the right elements.

```
npx playwright test --project chromium --debug structures
```

test-e2e/panels/structures.test.ts
hamed-musallam: @targos 
it is not guaranteed because the structure panel is reused in the prediction panel you should take care in case the prediction panel is opened

lpatiny: But in this testcase the prediction panel is not expected to be there so we prefer there is an error if suddenly there is a second structure editor.

test-e2e/panels/structures.test.ts
wadjih-bencheikh18: fixed thank you

test-e2e/panels/structures.test.ts
```diff
@@ -7,36 +7,105 @@ test('should draw structure and display it with MF', async ({ page }) => {
...
-  // The number of molecules should now be visible in the panel.
-  await expect(nmrium.page.locator('text=1 / 1 >> nth=0')).toBeVisible();
+    // The number of molecules should now be visible in the panel.
+    await expect(nmrium.page.locator('text=2 / 2')).toBeVisible();
+  });
+  await test.step('Switch between molecules', async () => {
+    // Start with the second molecule.
+    await expect(nmrium.page.locator('text=2 / 2')).toBeVisible();
+    await expect(nmrium.page.locator('text=C6H12 - 84.16')).toBeVisible();
+
```
lpatiny, line 93: Please also check there is a SVG

Issue #1694, opened 2022-08-11
Test case of shortcuts for processing


lpatiny:
https://user-images.githubusercontent.com/1484241/184194187-ec1fc1ae-aea8-4a5e-b8f2-bfba216664ab.mp4

wadjih-bencheikh18:
clicking A in the first step does not apply apodization filter, instead it activate phase correction filter

https://user-images.githubusercontent.com/56769876/184373211-a1b5a864-334e-4677-92ce-a01a9eb62e52.mp4

PR #1698, opened 2022-08-12
Test filters with shortcuts

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/1694


test-e2e/panels/filters.test.ts
```diff
@@ -33,22 +36,51 @@ async function fourierTransformFilter(nmrium: NmriumPage) {
...
 }
 
-async function phaseCorrectionFilter(nmrium: NmriumPage) {
-  await nmrium.clickTool('phaseCorrection');
-  await nmrium.page.fill('data-test-id=input-ph1', '-100');
-  await nmrium.page.fill('data-test-id=input-ph0', '-104');
-
-  // input debounce for 250ms
-  await nmrium.page.waitForTimeout(250);
+async function phaseCorrectionFilter(
```
lpatiny, line 39: @targos I would like your opinion about where to put those functions. If I understood well your idea is to move the code as a method of NmriumPage (just the action, not the expect) ?

The number of methods (and associated code) will become quite big so I would also suggest there is one file per big method of NmriumPage instance.

targos: I agree with that. In this case for example I would add a method named `applyPhaseCorrection(options)` to the class.

wadjih-bencheikh18: only applyPhaseCorrection functions? what about others?

targos: It's up to you. anything that may be done in more than one test can be there.

wadjih-bencheikh18: okey perfect thanks

wadjih-bencheikh18: > anything that may be done in more than one test can be there.

i added applyPhaseCorrection , in the next PR if i reused any other function i will add it in the class

Issue #1703, opened 2022-08-13
New testcase starting from 1H spectrum test


lpatiny:
We will do something very similar as the molecule test but staring from an existing file (that contains already a molecule).

https://user-images.githubusercontent.com/1484241/184478344-d8921f35-80f3-4cfd-b15a-596d3180f700.mp4

Check that when you copy to clipboard the second molecule, the text obtained has the correct number of lines (split('\n') and count the line) and that one of the line is `M  V30 COUNTS 6 6 0 0 0`

```

Actelion Java MolfileCreator 2.0

  0  0  0  0  0  0              0 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 6 6 0 0 0
M  V30 BEGIN ATOM
M  V30 1 C 13.5625 -16.125 0 0
M  V30 2 C 13.5625 -17.625 0 0
M  V30 3 C 14.8615 -18.375 0 0
M  V30 4 C 16.1605 -17.625 0 0
M  V30 5 C 16.1605 -16.125 0 0
M  V30 6 C 14.8615 -15.375 0 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 2 1 2
M  V30 2 1 2 3
M  V30 3 2 3 4
M  V30 4 1 4 5
M  V30 5 2 5 6
M  V30 6 1 6 1
M  V30 END BOND
M  V30 END CTAB
M  END
```

Check that you can add a third molecule and that there are the correct arrows (previous on the last molecule, prev / next for the middle molecule, next for the first molecule).

Check that if you delete the first molecule 2 are left and check the MF of the 2 molecules as well as the presence / absence of the next / prev arrows.

Check that you can have a 'floating' molecule and that there are 2 molecules displayed somewhere and disappear on delete.

PR #1704, opened 2022-08-13
1h spectrum molecules test

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/1703

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -33,6 +33,7 @@ const config: PlaywrightTestConfig = {
       use: {
         ...devices['Desktop Chrome'],
         viewport: viewportOverride,
+        permissions: ['clipboard-read', 'clipboard-write'],
       },
     },
     {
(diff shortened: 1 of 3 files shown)


lpatiny:
Please take care that we would like to Copy as molfile (and not save). 

After copy as molfile you may as well paste the molecule to check if there is really a new molecule inserted.

lpatiny:
When we add those structures in the 'structure' panel they should also appear in the dialog box in integrals and ranges. Please check that this is the case.

wadjih-bencheikh18:
> After copy as molfile you may as well paste the molecule to check if there is really a new molecule inserted.

Would i give chromium clipboard read access?

lpatiny:
> Would i give chromium clipboard read access?

If needed yes but this time is copy as text and maybe the rights are different.

PR #1709, opened 2022-08-15
chore: update OpenChemLib and related dependencies

targos:


targos:
@lpatiny Can you check that it works as you expect with OCL 8 ?

lpatiny:
Except for the color problem in the editor versus SVG it is ok (red becomes blue)

image


image

lpatiny:
The colors I expected ;)

Issue #1708, opened 2022-08-14
New test case: 2d analysis


lpatiny:
https://user-images.githubusercontent.com/1484241/184524274-bd190345-b0ab-4225-8c9f-d12798bdfcdb.mp4

PR #1711, opened 2022-08-17
2d analysis test

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/1708


lpatiny:
@wadjih-bencheikh18  In general when we check for a SVG element, could you add a condition on the length of this element (length greater than XXX). The goal is to check that if suddenly the SVG contains only one segment instead of a huge path we can detect it. Maybe as minimal length the current length of the element * 80%. You may also add this kind of condition in other tests you did.

Otherwise it looks good to me !

test-e2e/panels/spectra.test.ts
```diff
@@ -89,7 +89,7 @@ test('Check change spectrum color, Should be white', async ({ page }) => {
   await expect(whiteSpectrumLine).toBeVisible();
 });
 
-test('Should 2d deactivate spectrum', async ({ page }) => {
+test.skip('Should 2d deactivate spectrum', async ({ page }) => {
```
lpatiny, line 92: You forgot to remove the '.skip'. There is also the possibility to focus on a test using '.only'. Could be easier than adding skip everywhere.

wadjih-bencheikh18: yes sorry, thank youu

test-e2e/NmriumPage.ts
```diff
@@ -76,4 +76,13 @@ export default class NmriumPage {
       this.page.locator('_react=FilterPanel >> text=Phase correction'),
     ).toBeVisible();
   }
+  public async checkSVGLength() {
```
lpatiny, line 79: The idea here is still to hardcode in the test a minimal value and not base it on bodyLength.

This means that you need to first time to make a console.log to get the length of the SVG and then hardcode this value.

We could consider that the test pass if the value is >0.8 * expected value and <1.2 * expected value.

test-e2e/panels/spectra.test.ts
```diff
@@ -106,3 +106,257 @@ test('Should 2d deactivate spectrum', async ({ page }) => {
...
+    ).toBeVisible();
+    // Check that spectrum color changed
+    await expect(
+      nmrium.page.locator(
+        'data-test-id=spectrum-line >> _react=Line[display.color="#ddb1c9ff"]',
+      ),
+    ).toBeVisible();
+
+    // Check svg
+    await nmrium.checkSVGLength();
```
lpatiny, line 159: Need to add an hardcoded value there

wadjih-bencheikh18: okey i get it thanks

PR #1715, opened 2022-08-19
Fix and improve DraggableStructure

### LINKED ISSUE #1702: Floating molecule and local state ?
There is an issue with floating molecule. You can click twice to edit the molecule from the floating molecule which is nice but while the molecule is correctly changed in the panel, the floating molecule is not changed (local state ???).

Each molecule must have a unique ID and the floating molecule should only have this uniqueID as property. If the molecule is deleted this uniqueID will disappear and the floating molecule will disappear as well. If the molecule is edited it should be refreshed automatically.

This should be done after: https://github.com/cheminfo/nmrium/issues/1700

stropitek:
Closes #1702


stropitek:
@wadjih-bencheikh18 can you take over this PR and work on the state refactor:

Create new property `view.floatingMolecules` in the main state of the application (which comes from `ChartContext`)

The new state is structured like this:
```ts
Array<{
  id: string; // reference to the id in `molecules`
  visible: boolean; // If the floating molecule is shown
}>
```

and it should replace the `isFloat` state in `molecules`

wadjih-bencheikh18:
@stropitek i don't have access to push in this branch, do you want me to create new PR?

stropitek:
> i don't have access to push in this branch, do you want me to create new PR?

Yes please do so

wadjih-bencheikh18:
> Yes please do so

new PR created you can delete this PR

src/component/reducer/Reducer.ts
```diff
@@ -85,6 +85,7 @@ export const getInitialState = (): State => ({
   activeSpectrum: null,
   mode: 'RTL',
   molecules: [],
+  floatingMolecules: [],
```
lpatiny, line 88: @stropitek I think this property should not be at the first level but rather in another prop 'view' or something like that. WDYT ?

lpatiny:
The functionality is ok. If we edit the molecule it correctly change the floating molecule as well. I think however there is a level missing in the state to store all the 'view' related information.

src/component/reducer/Reducer.ts
stropitek: Let's but in a `view` property, it will help identify what has already been identified as not belonging to the core data state.

stropitek:
Good work @wadjih-bencheikh18 👍🏻 

I can't really tell what changed in the tests from this but the implementation LGTM

stropitek:
Just change the datastructure to add the `view` level and we can merge this

wadjih-bencheikh18:
> change the datastructure to add the `view` level

Could you please explain this, Do you mean NMRiumData ?

stropitek:
> Could you please explain this, Do you mean NMRiumData ?

No I'm referring to the unresolved conversation: https://github.com/cheminfo/nmrium/pull/1715#discussion_r953680900

It just that `floatingMolecules` should not be at the first level of the state, but in a `view` property.

lpatiny:
there is still a problem with the toggle button to display the molecule. When we close the molecule the button does toggle.

wadjih-bencheikh18:
> When we close the molecule the button does toggle.

I will fix it now and add the test case,
Thank you

wadjih-bencheikh18:
I added new element named `ActiveButton` (not sure about the name) that allow toggle btn and the style change depend `value` prop

Issue #1658, opened 2022-07-29
Highlight atom on floating windows


lpatiny:
There is an issue in the highlight of floating molecules. When you go over an assigned atom it is not highlighted in the other molecule.

![2022-07-29 09 52 25](https://user-images.githubusercontent.com/1484241/181711750-e110972c-6ed1-4e98-ac47-6b42e350ccfc.gif)

lpatiny:
@wadjih-bencheikh18 Could you have a look if there is something obvious ? We have a highlight system in NMRium that allows to highlight various element in the interface. What is currently not working is the 'auto' highlight. If you highlight an atom in a molecule I would expect that it is being highlighted in all the molecules.

PR #1724, opened 2022-08-25
Sync highlight molecules

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1658


wadjih-bencheikh18:
To sync highlight i accepted `HighlightedSource.UNKNOWN` , maybe the right way is to add new `HighlightedSource` for this case

lpatiny:
But what is listening the 'Range' ? Because when we go over an atom it highlights the corresponding range. I would expect that there is when we go over an atom a `HiglightedSource.ATOM`

I'm also surprised about `HighlightedSource` name. If it is the source of the highlight event I would rather call it `HighlightSource` or `HighlightEventSource` (even better). But it should be changed if it is really the source of the event.

Issue #1720, opened 2022-08-25
Keep floating molecule coordinates in state


stropitek:
Keep the floating molecule's coordinates in the `floatingMolecules` state.

The state should be updated at the end of the drag and drop action. The draggable svg component already has a callback prop to handle this event, so it should be pretty straightforward.

This will prevent the floating molecule to reset its state when the component remounts.

PR #1730, opened 2022-08-26
refactor: add floating molecule position in state

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1720


lpatiny:
It works correctly if you have only 1D spectra but in case you have a 2D spectrum the position is reset.

wadjih-bencheikh18:
> It works correctly if you have only 1D spectra but in case you have a 2D spectrum the position is reset. ![2022-08-26 17 13 21]


Fixed thank you

lpatiny (approved):
Functionality is ok.

@stropitek Could you check the implementation and merge this branch ?

stropitek:
> Could you check the implementation and merge this branch ?

You can use the "request review" feature instead of pinging me ;-)

src/component/reducer/actions/MoleculeActions.ts
```diff
@@ -106,7 +107,22 @@ function floatMoleculeOverSpectrum(draft: Draft<State>, action) {
...
+    draft.view.floatingMolecules.push({
+      id,
+      visible: true,
+      position: DRAGGABLE_STRUCTURE_INITIAL_POSITION,
+    });
+  }
+}
+function changeFloatMoleculePosition(draft: Draft<State>, action) {
+  const { id, position } = action.payload;
+  const moleculeIndex = draft.view.floatingMolecules.findIndex(
```
stropitek, line 119: Please prefer `find` over `findIndex` in this case

src/component/tool/FloatMoleculeStructures/index.tsx
```diff
@@ -47,6 +47,7 @@ export function FloatMoleculeStructuresInner(
               ranges,
               activeTab,
               displayerMode,
+              position: molecule.position,
               molecule: molecules.find(
```
stropitek, line 51: Not in the scope of this PR but just to mention one of our "best practices":

Instead of using `as`, create a variable and `assert` it is defined since it should be an error for the molecule not to exist.

I don't know if nmrium already has some typescript-friendly assert utility functions but I can provide them if needed.

If one day the error happens, it will be easy to know exactly where the error stems from from the stack trace

wadjih-bencheikh18: Thank you
it's fixed now

stropitek:
You can use this little utility function for the assertion:

```ts
export function assert(value: unknown, message?: string): asserts value {
  if (!value) {
    throw new Error(`unreachable${message ? `: ${message}` : ''}`);
  }
}
```

Issue #1688, opened 2022-08-11
FFT of data that is not power of 2


lpatiny:
Currently if data is not a power of 2 the fft fails with a message 'Something went wrong'.

I think by default we can use the next power of 2.

[(using next-power-of-two)](https://github.com/mattdesl/next-power-of-two)

I find this kind of code amazing: https://github.com/mattdesl/next-power-of-two/blob/master/index.js

PR #1732, opened 2022-08-27
fix: peaks, signals, ranges, zones with id

jobo322:
close: #1688


src/data/data1d/filter1d/fft.ts
```diff
@@ -60,3 +69,15 @@ function generateXAxis(datum1D) {
   }
   return xAxis;
 }
+
+function isPowerOfTwo(n) {
+  return n !== 0 && (n & (n - 1)) === 0;
+}
+
+function hasSameLength(data: Data1D) {
+  const xLength = data.x.length;
```
lpatiny, line 78: Why using a loop (and even a function) just to do:

(data.x.length === data.re.length) && (data.x.length === data.im.length)

or did I miss something.

src/data/data1d/filter1d/utils/padDataToNextPowerOfTwo.ts
```diff
@@ -0,0 +1,48 @@
+import { xSequentialFill } from 'ml-spectra-processing';
+
+import { Datum1D } from '../../../types/data1d';
+
+export function padDataToNextPowerOfTwo(
+  datum1D: Datum1D,
+  digitalFilterApplied: boolean,
```
lpatiny, line 7: There is something I don't understand here. When is the digital filter applied and how does it impact this code. Because in the list of filters it appears at the beginning and I would not expect that it impacts further steps.

image

package.json
```diff
@@ -70,13 +70,13 @@
...
     "ml-conrec": "^3.2.1",
-    "ml-gsd": "^12.0.0",
+    "ml-gsd": "12.1.0",
     "ml-spectra-processing": "^11.12.0",
     "ml-stat": "^1.3.3",
     "multiplet-analysis": "^2.0.0",
     "nmr-correlation": "^2.3.3",
     "nmr-parser": "^1.9.1",
-    "nmr-processing": "^9.0.3",
+    "nmr-processing": "^9.0.3-pre.1661576806",
```
targos, line 79: Just making sure this isn't merged with the preversion.

package.json
```diff
@@ -70,13 +70,13 @@
     "ml-array-xy-equally-spaced": "^1.2.1",
     "ml-baseline-correction-regression": "^1.0.0",
     "ml-conrec": "^3.2.1",
-    "ml-gsd": "^12.0.0",
+    "ml-gsd": "12.1.0",
```
targos, line 73: @jobo322 Can you explain how you update dependencies? I don't understand why the caret is removed so often in your updates.

src/component/1d/peaks/PeaksShapes.tsx
```diff
@@ -71,6 +71,7 @@ function PeaksShapesSum(props: { vAlign: number }) {
 
   return (
     <path
+      data-test-id="peaks-shapes-sum"
```
targos, line 74: Why add a test id here?

jobo322: I have not idea why it is there

package.json
jobo322: I use `npm i [packageName]@[version]`, I don't use this caret in this case because I was not sure about the ml-gsd version, it was intentional and I did returns the caret symbol.

src/component/1d/peaks/PeaksShapes.tsx
jobo322: I will remove it

package.json
jobo322: @lpatiny did you test the behavior of nmrium with this pre-release of nmr-processing? can I merge the PR in nmr-processing?

package.json
lpatiny: I can not reproduce this. The caret is not removed when I do `npm i ml-gsd@12.1.0` For you Alejandro it is being removed ? 

There is an option in npm that allows to force the version: https://stackoverflow.com/questions/30656189/make-npm-install-save-add-a-strict-version-to-package-json

Could you do `npm config ls -l` and check that `save-exact = false`

package.json
lpatiny: I didn't see any new issue with this version.

src/data/data1d/filter1d/fft.ts
jobo322: yes, I remove the function and did it directly as the conditional

package.json
jobo322: yes, exact version flag is false, the use of --save-exact is the best way as Michael let me know to install a pre-release

lpatiny: So did you do `npm i --save-exact ml-gsd@latest` ?

How did you do to remove this `^`

jobo322: No, this is the current way to install a fixed version. I remove the caret manually for ml-gsd. I did it because I was confused, as to why I got installed a different version, for nmr-processing I got installed always version 12.0.0, so in order to test the current state of the code I used a fixed version of ml-gsd to be sure.

I used this fixed version as we use the fixed version of pre-release. Now I understand that npm prefers the stable version instead of a pre-release version, it was the source of my confusion.

I catch the idea that we always use the caret from the above comments, I will only use the fixed version for pre-release versions

Issue #1723, opened 2022-08-25
Refactor active spectrum


lpatiny:
We need to add a new property in 'view' called 'spectra'

This new `spectra` property will be an object with the property:
- currentSpectrumId

https://github.com/cheminfo/nmrium/blob/84f399fdf053fa866a554091b72b3c408982e34e/src/component/hooks/useSpectrum.ts#L7-L18

lpatiny:
Currently there can be only one active spectrum at the same time.

You can change from one category of spectra to another. Each category keeps his currently selected spectrum.

![2022-08-25 14 40 27](https://user-images.githubusercontent.com/1484241/186667258-8b91a3c0-fc9c-4102-9e35-f17f7a12a723.gif)

@wadjih-bencheikh18 tell us if you see some problem in this refactoring as it can have side effects that I currently don't see.

wadjih-bencheikh18:
> Currently there can be only one active spectrum at the same time.

I didn't undersand what do you mean in this comment.
there's only one active spectrum but each tab active spectrum is saved in `tabActiveSpectrum`

lpatiny:
I added some comments in the PR

PR #1739, opened 2022-08-31
refactor: add spectra view state

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1723


src/component/reducer/actions/SpectrumsActions.ts
```diff
@@ -116,6 +116,7 @@ function handleChangeActiveSpectrum(draft: Draft<State>, activeSpectrum) {
     }
     activeSpectrum = { ...activeSpectrum, index: newIndex };
     draft.activeSpectrum = activeSpectrum;
+    draft.view.spectra.currentSpectrumId = activeSpectrum.id;
     draft.tabActiveSpectrum[draft.activeTab] = activeSpectrum;
```
lpatiny, line 104: Using the activeTab is important because for each tab we have an active spectrum. So I was wrong in my issue.

Currently it surprisingly works but I don't understand why ... This `spectra` property should contain a property `activeSpectra` that is an object that contains one property per tab. So all the changes must depend on `activeTab`

lpatiny (changes requested):

src/component/reducer/actions/SpectrumsActions.ts
wadjih-bencheikh18: in the project we use `tabActiveSpectrum` to save the active spectrum for each tab, so
do you want me to add tab active spectrum and active tab in the view?

lpatiny: Thanks for checking this. Then I don't know where is currentSpectrumId (is it used ?)

I would expect that 'tabActiveSpectrum' goes indeed in

view.spectra.activeSpectraId[activeTab]

We should check if we need to keep the activeSpectrum or just it's id. This may require a lot of changes so we will currently wait with this PR because there is another one (load-save) that changes a lot of code.

wadjih-bencheikh18: we need the index in `hasAcceptedSpectrum` function in `ToolsActions`

src/component/1d/jCouplingGraph/JGraph.tsx
```diff
@@ -42,7 +42,9 @@ export default function JGraph() {
     activeSpectrum,
     height,
     view: { ranges: rangeState },
-    activeTab,
+    view: {
+      spectra: { activeTab },
+    },
```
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed thank you

lpatiny:
@wadjih-bencheikh18 could you check why some test fail ?

Otherwise the functionality looks ok to me.

wadjih-bencheikh18:
> @wadjih-bencheikh18 could you check why some test fail ?
> 
> Otherwise the functionality looks ok to me.

Yes i'm working on it
Thank you

src/component/reducer/actions/DomainActions.ts
```diff
@@ -73,18 +74,23 @@ function getDomain(drfat: Draft<State>) {
...
+      spectra: { activeSpectra, activeTab },
+    },
+    data,
+  } = state;
 
   const nucleus = activeTab.split(',');
 
   try {
-    xArray = data.reduce((acc, datum: Datum1D | Datum2D) => {
+    xArray = data.reduce((acc: any, datum: Datum1D | Datum2D) => {
```
stropitek, line 93: Can you fix this use of `any`? `reduce` is a generic that takes the type of the accumulated value:
`data.reduce<Array<[number, number]>>(...)`

src/component/reducer/actions/DomainActions.ts
```diff
@@ -95,7 +101,7 @@ function get2DDomain(state) {
       return acc;
     }, []);
 
-    yArray = data.reduce((acc, datum: Datum1D | Datum2D) => {
+    yArray = data.reduce((acc: any, datum: Datum1D | Datum2D) => {
```
stropitek, line 104: ditto

src/component/reducer/actions/DomainActions.ts
```diff
@@ -110,9 +116,9 @@ function get2DDomain(state) {
     console.log(e);
   }
 
-  const spectrumsIDs = nucleus.map((n) => tabActiveSpectrum[n]?.id);
+  const spectrumsIDs = nucleus.map((n) => activeSpectra[n]?.id);
 
-  const filteredData = data.reduce((acc, datum: Datum1D | Datum2D) => {
+  const filteredData = data.reduce((acc: any, datum: Datum1D | Datum2D) => {
```
stropitek, line 121: ditto

src/component/reducer/actions/FiltersActions.ts
```diff
@@ -29,8 +29,11 @@ import { resetSelectedTool } from './ToolsActions';
 
 function shiftSpectrumAlongXAxis(draft: Draft<State>, shiftValue) {
   //apply filter into the spectrum
-  if (draft.activeSpectrum?.id) {
-    const index = draft.activeSpectrum?.index;
+
+  const activeSpectrum =
+    draft.view.spectra.activeSpectra[draft.view.spectra.activeTab];
+  if (activeSpectrum?.id) {
+    const index = activeSpectrum?.index;
```
stropitek, line 36: Can you implement a helper which does this as it's expected to be used in a lot of places within the reducer.

Also I think this should throw instead of doing nothing if no active spectrum exists. The application should never call this reducer if no active spectrum exists.

Something like a `getActiveSpectrumOrFail(draft)` helper

src/component/reducer/actions/FiltersActions.ts
```diff
@@ -42,8 +45,10 @@ function shiftSpectrumAlongXAxis(draft: Draft<State>, shiftValue) {
 }
 
 function applyZeroFillingFilter(draft: Draft<State>, action) {
-  if (draft.activeSpectrum?.id) {
-    const index = draft.activeSpectrum.index;
+  const activeSpectrum =
+    draft.view.spectra.activeSpectra[draft.view.spectra.activeTab];
+  if (activeSpectrum?.id) {
```
stropitek, line 50: ditto

src/component/reducer/actions/FiltersActions.ts
```diff
@@ -59,8 +64,10 @@ function applyZeroFillingFilter(draft: Draft<State>, action) {
 }
 
 function calculateApodizationFilter(draft: Draft<State>, action) {
-  if (draft.activeSpectrum) {
-    const index = draft.activeSpectrum.index;
+  const activeSpectrum =
+    draft.view.spectra.activeSpectra[draft.view.spectra.activeTab];
+  if (activeSpectrum) {
```
stropitek, line 69: ditto (I'll stop the dittos here because it's everywhere)

src/component/reducer/actions/ToolsActions.ts
```diff
@@ -57,23 +57,25 @@ function resetSelectedTool(draft: Draft<State>, filterOnly = false) {
...
+    draft.view.spectra.activeSpectra[draft.view.spectra.activeTab];
   if (draft?.data.length > 0) {
     if (selectedTool) {
       // start Range edit mode
       if (selectedTool === options.editRange.id) {
         // TODO: use assert
-        if (draft.activeSpectrum === null) {
+        if (activeSpectrum === null) {
           throw new Error('Spectrum must have id');
         }
```
stropitek, line 69: you can apply the TODO

lpatiny:
We will still need to merge this PR https://github.com/cheminfo/nmrium/pull/1753 before merging this one.

lpatiny (approved):
The functionalities are ok to me and spectra

@stropitek if the code is ok could you merge this PR ?

Issue #1722, opened 2022-08-25
Toggle zones


lpatiny:
This should be done after: https://github.com/cheminfo/nmrium/issues/1721

For 2D spectra we don't have ranges but 'zones'.

If you load the 'full ethylbenzene' example you are able to add 'zones' and toggle different properties of the zone (you need to zoom quite a lot to check the difference of those toggles'.

Again those 3 properties (show or hide) should be in the 'view' context. It is also related to the currently displayed spectrum. A new property in the view context names `zones` should be added. It contains an array of:
- spectrumId
- showZones
- showSignals
- showPeaks

PR #1741, opened 2022-08-31
refactor: add zone state view

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1722


src/component/panels/ZonesPanel/ZonesPanel.tsx
```diff
@@ -154,9 +170,15 @@ function ZonesPanelInner({ zones, activeTab, xDomain, yDomain, experiment }) {
     setFlipStatus(false);
   }, []);
 
-  const visibilityHandler = useCallback((key) => {
-    Events.emit('onZonesVisibilityChange', { key });
-  }, []);
+  const handleSetShowZones = useCallback(() => {
```
stropitek, line 173: using useCallback or useMemo for things that are passed to a component that is not memoized is actually a (very small) decrease in performance

In general our approach to performance is do not worry about it too much at first unless there are heavy computations involved, or you are working with state (typically in a react context) which trigger the re-rendering of the whole page.

At some point we will refocus on performance looking at bottlenecks, but it is not a priority today.

src/component/panels/ZonesPanel/ZonesPanel.tsx
```diff
@@ -258,18 +280,26 @@ const MemoizedZonesPanel = memo(ZonesPanelInner);
...
-
+  const {
+    displayerKey,
+    xDomain,
+    yDomain,
+    activeTab,
+    view: { zones: zoneState },
+  } = useChartData();
+  const { zones, info, id } = useSpectrum(emptyData) as Datum2D;
+  const btnState = zoneState.find((r) => r.spectrumID === id) || zoneStateInit;
```
stropitek, line 291: why `btnState`? and not something like `zoneProps`?

src/component/panels/ZonesPanel/ZonesPanel.tsx
wadjih-bencheikh18: Do you want me to revert states changes ?

stropitek: No, just remove the 3 next `useCallback` as you can just declare regular functions without hurting performance

stropitek: I invite you to read
https://www.joshwcomeau.com/react/why-react-re-renders/
https://www.joshwcomeau.com/react/usememo-and-usecallback/

wadjih-bencheikh18: > just remove the 3 next `useCallback`

Done,fixed


> I invite you to read

Thank you

stropitek (approved):
LGTM after conflicts are resolved

Issue #1742, opened 2022-09-01
onViewChange


lpatiny:
NMRium component allows the property 'onDataChange' that is called whenever the data is changed.

We would like a similar property `onViewChange` that will be triggered when the property `view` is changing.

In order to debug callbacks we can use this special tab:

![image](https://user-images.githubusercontent.com/1484241/187906188-41a8e133-be15-46bc-84d2-411811657faa.png)

And on the right part we could add the 'onViewChange' result in another JSON.

In order to see when the data or view is changing it would be nice if there is a kind of visual effect showing that the value has been updated.

The goal of this view is to check if we don't have unnecessary or missing callbacks

lpatiny:
In order to debug you can drag / drop the following decompressed file

[1h.jdx.zip](https://github.com/cheminfo/nmrium/files/9471036/1h.jdx.zip)

PR #1743, opened 2022-09-01
Implement onViewChange callback

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1742


src/component/reducer/IgnoreActions.ts
```diff
@@ -34,11 +34,23 @@ const ignoreActions = [
...
 
-export default checkActionType;
+const viewChangeActions = [
+  types.CHANGE_FLOAT_MOLECULE_POSITION,
+  types.FLOAT_MOLECULE_OVER_SPECTRUM,
+  types.SHOW_J_GRAPH,
+  types.SHOW_MULTIPLICTY_TREES,
+  types.SHOW_RANGES_INTEGRALS,
+];
+
```
lpatiny, line 52: You should create a new useEffect that depends only on the viewState so that we don't need to specify the list of Actions.

The work you are currently doing (separation of view state from data state) is being done exactly to avoid this hack.

lpatiny (changes requested):
Please don't forget also to add a way to test the functionality in the 'Test' tab as described in the issue.

image

src/component/reducer/IgnoreActions.ts
wadjih-bencheikh18: thank you

wadjih-bencheikh18: i used useEffect now instead of check Actions

lpatiny:
@wadjih-bencheikh18 Could you still make a visual effect when the callback is called ?

image


I would like to have a flash around the component for example when the data or view callback is called.

wadjih-bencheikh18:
> Could you still make a visual effect when the callback is called ?

Effect added.

lpatiny (approved):

lpatiny:
Functionality is ok for me. @stropitek Please merge it if the code is ok.

src/component/NMRium.tsx
```diff
@@ -217,6 +220,10 @@ function InnerNMRium({
     }
   }, [onDataChange, state]);
 
+  useEffect(() => {
+    onViewChange?.(state.view);
+  }, [onViewChange, state.view]);
```
stropitek, line 225: You can remove the `onViewChange` dependency and put it in a ref instead.

If the api consumer redeclares the callback on every render, we don't want the callback to be called without any view state change.

src/demo/views/Test.tsx
```diff
@@ -78,14 +78,27 @@ export default function Test() {
       }
     })();
   }, []);
+  const viewChangeHandler = useCallback((data) => {
```
stropitek, line 81: The `useCallback` can be removed after you make the change with the ref

src/component/NMRium.tsx
```diff
@@ -217,6 +221,10 @@ function InnerNMRium({
     }
   }, [onDataChange, state]);
 
+  useEffect(() => {
+    onViewChange.current?.(state.view);
+  }, [state.view]);
```
targos, line 226: @stropitek The ref is never updated, so if the callback changes, we won't be calling the new version.

PR #1753, opened 2022-09-05
use nmr-load-save

jobo322:


lpatiny (approved):
The functionality is ok and a lot of code not related to graphical interface has been moved to nmr-load-save which is great.

@stropitek do you see anything shocking in the code ?

src/component/loader/DropZone.tsx
```diff
@@ -68,188 +50,26 @@ function DropZone(props) {
...
-                'The file extension must be zip, dx, jdx, json, mol, nmredata or nmrium.',
-              );
-          }
-        } catch (e: any) {
-          dispatch({ type: SET_LOADING_FLAG, isLoading: false });
-          alert.error(e.message);
-        }
+    (files) => {
+      files.forEach((file) => {
+        Object.defineProperty(file, 'webkitRelativePath', { value: file.path });
```
stropitek, line 56: Why is this property added? For browser interoperability?

If there are some browser specificities, shouldn't it be up to the library to handle them?

@lpatiny

src/component/loader/DropZone.tsx
```diff
@@ -68,188 +50,26 @@ function DropZone(props) {
...
+      });
+
+      try {
+        dispatch({ type: LOAD_DROP_FILES, files });
```
stropitek, line 60: the async that is run inside the dispatch will not be caught in this `try...catch` statement. The error logic should move inside the action dispatcher.

src/component/reducer/Reducer.ts
```diff
@@ -457,33 +456,11 @@ export function dispatchMiddleware(dispatch) {
...
-        }
-        break;
-      }
-      case types.LOAD_NMREDATA_FILE: {
-        void nmredataToNmrium(action.file, usedColors).then((data) => {
-          action.payload = Object.assign(data, { usedColors });
+      case types.LOAD_DROP_FILES: {
+        const { files } = action;
+        action.usedColors = usedColors;
+        void readDropFiles(files, usedColors).then((data) => {
```
stropitek, line 462: The error should be handled here

src/component/loader/DropZone.tsx
jobo322: This property is needed for brukerconverter, the module react-dropzone does not fill `webkitRelativePath`. it is a temporal hack wheel the react-dropzone is reemplaced for another solution.

lpatiny: Yes it is a 'validated' hack ;)

src/component/loader/DropZone.tsx
jobo322: I did move the try...catch block to Reducer.

package.json
```diff
@@ -74,7 +74,8 @@
     "ml-stat": "^1.3.3",
     "multiplet-analysis": "^2.0.0",
     "nmr-correlation": "^2.3.3",
-    "nmr-parser": "^1.9.1",
+    "nmr-load-save": "0.1.0-pre.1662845670",
+    "nmr-parser": "2.0.0-pre.1662404586",
```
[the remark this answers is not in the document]
jobo322: all of package are using a stable release

src/component/reducer/Reducer.ts
```diff
@@ -457,37 +455,20 @@ export function dispatchMiddleware(dispatch) {
...
-            action.payload = { data, usedColors };
+      case types.LOAD_DROP_FILES: {
+        const { files } = action;
+        action.usedColors = usedColors;
+        try {
+          void readDropFiles(files, usedColors).then((data) => {
+            action.data = data;
             dispatch(action);
           });
+        } catch (e: any) {
```
stropitek, line 466: You cannot `try...catch` an async error unless you use `await`.

Here you can't do that because it's not an async function. So handle your error in `readDropFiles(...).catch(...)`

Please test that error handling is working properly by throwing a random error inside the `then` handler.

src/component/reducer/Reducer.ts
```diff
@@ -457,35 +455,18 @@ export function dispatchMiddleware(dispatch) {
...
+        const { files } = action;
+        action.usedColors = usedColors;
+        void readDropFiles(files, usedColors)
+          .then((data) => {
+            action.data = data;
             dispatch(action);
+          })
+          .catch((e: any) => {
+            dispatch({ type: types.SET_LOADING_FLAG, isLoading: false });
+            throw new Error(e);
```
stropitek, line 468:
```suggestion
            reportError(e);
```

stropitek: @jobo322 you are creating an uncaught exception by throwing here

src/component/reducer/Reducer.ts
```diff
@@ -457,35 +455,18 @@ export function dispatchMiddleware(dispatch) {
...
-          void SpectraManager.addBruker(
-            { display: { name: zipFile.name } },
-            zipFile.binary,
-            usedColors,
-          ).then((data) => {
-            action.payload = { data, usedColors };
+      case types.LOAD_DROP_FILES: {
+        const { files } = action;
+        action.usedColors = usedColors;
+        void readDropFiles(files, usedColors)
```
stropitek, line 461:
```suggestion
       readDropFiles(files, usedColors)
```

Issue #1706, opened 2022-08-14
Position of dialog boxes


lpatiny:
If you resize the window (reduce the size) the position of the dialog boxes are based on the window original size and not the current size of the window. This seems the case for all the temporary components like the tip that appear when you do CTRL+C for copy the image (you need a spectrum to test this function).

This can easily be tested by reducing the size of the window and click on general preferences.

![2022-08-14 07 28 12](https://user-images.githubusercontent.com/1484241/184523948-c9a472e4-d793-4c4f-af68-f1b368eb36c2.gif)

PR #1757, opened 2022-09-08
fix dialog box position

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1706

### DIFF
--- a/src/component/elements/popup/Modal/Provider.tsx
+++ b/src/component/elements/popup/Modal/Provider.tsx
@@ -78,17 +78,12 @@ function Provider({
     [setModal],
   );
   // todo: optimize this
-  const wrapperRect = wrapperRef?.getBoundingClientRect();
-  const parentStyle = useMemo(() => {
-    return (
-      wrapperRect || {
-        top: 0,
-        left: 0,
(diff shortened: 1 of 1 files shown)


src/component/elements/popup/Modal/Provider.tsx
```diff
@@ -77,17 +77,18 @@ function Provider({
     },
     [setModal],
   );
-
+  // todo: optimize this
+  const wrapperRect = wrapperRef?.getBoundingClientRect();
   const parentStyle = useMemo(() => {
```
targos, line 82: I'm pretty sure you can remove the `useMemo`, because `wrapperRect` is always a different object.

wadjih-bencheikh18: I added wrapperRect, because using wrapperRef as dependency doesn't work correctly (the prev version)
and i found this solution
https://gist.github.com/morajabi/523d7a642d8c0a2f71fcfa0d8b3d2846

lpatiny:
@targos The functionality is ok so if the code is ok for you please merge this PR.

PR #1767, opened 2022-09-23
fix: peak shape drawing

jobo322:


src/component/1d/peaks/usePeakShapesPath.ts
```diff
@@ -30,6 +30,9 @@ export function usePeakShapesPath(spectrum: Datum1D) {
       case 'peakShape':
         pathSeries = peakToXY(options.peak, {
           frequency: spectrum.info.originFrequency,
+          nbPoints: 1024,
+          from: options.peak.x - 0.03,
```
lpatiny, line 34: I don't like that much this magic value because this will fail when we are using another nucleus. Could you rather find a magic value :D that is based on the peak.width ?

lpatiny (approved):

src/component/1d/peaks/usePeakShapesPath.ts
jobo322: yes, I did use 9 times of the peak.width, seems to be a good range for Lorentzian shape

Issue #1783, opened 2022-09-29
Zoom and release outside 


lpatiny:
Currently the zoom is not tracking the mouse when it leaves the component.

![2022-09-29 15 13 15](https://user-images.githubusercontent.com/1484241/193040987-479fbd72-72f7-4379-8164-0cffae6d5168.gif)

Michael solved this problem in the split pane for example.

![2022-09-29 15 12 02](https://user-images.githubusercontent.com/1484241/193041013-f94a00c9-da2b-4271-b160-013497e0731a.gif)

This issue can be solved using an approach similar to https://github.com/zakodium-oss/analysis-ui-components/blob/901c6b15391ae030f0dc9e2c5795af59d476fd1a/src/components/split-pane/useSplitPaneSize.tsx#L124-L132

PR #1794, opened 2022-10-04
feat: improve brush tracker

hamed-musallam:


hamed-musallam:
@lpatiny 

this going to solve the problem of the Zoom and release outside  ( #1783 ), can you test it?

lpatiny (approved):
Functionnality is ok for me

PR #1816, opened 2022-10-14
support for inline jcamp from nmr-load-save

jobo322:
use readNMRiumObject instead fromJSON


src/component/reducer/Reducer.ts
```diff
@@ -456,9 +459,9 @@ export function dispatchMiddleware(dispatch) {
     switch (action.type) {
       case types.INITIATE: {
         if (action.payload) {
-          const { spectra, ...res } = migrate(action.payload);
-          void SpectraManager.fromJSON(spectra, usedColors).then((data) => {
-            action.payload = { spectra: data, ...res, usedColors };
+          const result = migrate(action.payload);
```
hamed-musallam, line 462: @jobo322 

the migration is longer need it here because you do the same in https://github.com/cheminfo/nmr-load-save/blob/b17d8af9a8ed22a0115a2e3f9e5f5f3bec03f9c1/src/reader/readNMRiumObject.ts#L17

jobo322: ups, many thanks

package.json
```diff
@@ -75,12 +75,13 @@
...
     "multiplet-analysis": "^2.0.0",
     "nmr-correlation": "^2.3.3",
-    "nmr-load-save": "^0.4.2",
+    "nmr-load-save": "0.4.2-pre.1666322459",
     "nmr-processing": "^9.2.0",
     "nmredata": "^0.9.0",
     "numeral": "^2.0.6",
     "openchemlib": "^8.0.1",
     "openchemlib-utils": "^2.0.0",
+    "playwright": "^1.27.1",
```
targos, line 84: ?

jobo322: yes, it was a mistake, I forgot to re-install It globally, then it write the package.json

targos: you shouldn't need to install it globally. I never did it personally.

targos: In general, manually installing global npm packages is discouraged since `npx` exists.

package.json
```diff
@@ -129,4 +128,4 @@
     "typescript": "~4.8.4",
     "vite": "^3.1.7"
   }
-}
+}
```
targos, line 131:
```suggestion
}

```

test-e2e/panels/ranges.test.ts
```diff
@@ -235,56 +235,11 @@ test('Range state', async ({ page }) => {
       await nmrium.page.locator('_react=MultiplicityTree').count(),
     ).toBeGreaterThan(0);
   });
-  await test.step('Change active spectrum', async () => {
```
targos, line 238: Why did you remove a test?

jobo322: I did remove the testing over a FID using auto-ranges picking. is it?

lpatiny: Actually @jobo322 we should test that the button Auto range picking does not existing in the page anymore. You could also check that the Apodization button is there.

It is indeed weird that this test was passing because the button Auto range picking after clicking on the first spectrum should not be there anymore.

PR #1818, opened 2022-10-15
test: fix test

hamed-musallam:


test-e2e/panels/ranges.test.ts
```diff
@@ -116,11 +116,11 @@ test('Automatic ranges detection should work', async ({ page }) => {
...
-    { s: '1.75 - 1.79', r: '0.49' },
-    { s: '1.95', r: '2.05' },
-    { s: '1.94', r: '2.05' },
-    { s: '2.15', r: '0.07' },
-    { s: '2.31 - 2.34', r: '1.01' },
+    { s: '1.75 - 1.79', r: '3.49' },
+    { s: '1.95', r: '14.67' },
+    { s: '1.94', r: '14.67' },
+    { s: '2.15', r: '0.53' },
+    { s: '2.31 - 2.34', r: '7.24' },
```
lpatiny, line 123: We should not fixed the test bug fix the bug that has introduced this changes.

Before we were automatically setting has sum for the ranges the number of H that were present in the molfile. This is still working correctly in the last release that is currently working on www.nmrium.org.

But in this release the sum is not automatically calculated. @hamed-musallam this may have been introduced by the changes of @jobo322. Please check were it should be corrected. In nmr-load-save ?

test-e2e/panels/spectra.test.ts
```diff
@@ -127,33 +127,35 @@ test('2d spectrum', async ({ page }) => {
     await expect(Tabs.last()).toHaveText('1H,1H');
   });
   await test.step('Change 1H spectrum color', async () => {
+    await nmrium.page.click('_react=InternalTab[tabid="1H"]');
+
     // Check ColorIndicator initial color
     await expect(
-      nmrium.page.locator('_react=ColorIndicator[color.color="#7c2353"]'),
+      nmrium.page.locator('_react=ColorIndicator[color.color="#C10020" i]'),
```
lpatiny, line 134: We also need to understand why this color changed. Moreover it is only composed of 5 symbols and color should be 6.

I would expect that we correct the code rather than fixing the test.

lpatiny (changes requested):
I have this impression that the test should not be changed but we need to understand exactly why those changes appeared and fix the root cause.

test-e2e/panels/spectra.test.ts
hamed-musallam: @lpatiny 

the first color should be always red #C10020, i do not know how it was working before !!!!!!!

lpatiny: Ok. And you know why there is not this `i` in `="#C10020" i]'),` ?

hamed-musallam: I just add it in case it is in a small or capital case. no matter how I enter the value.

test-e2e/panels/ranges.test.ts
hamed-musallam: the default value from nmr-load-save is 100

https://github.com/cheminfo/nmr-load-save/blob/08dcd987ab042d5d1de62a5512dd5cffbb930755/src/utilities/formatSpectrum1D.ts#L74-L79

hamed-musallam: all these options should be handed by NMRium and not by the nmr-load-save package. what do you think?

hamed-musallam: to make it work the sumAuto: true option should be in the ranges options.

in NMRium Once we initiate the 1D spectrum we initiate the ranges object with the default options if it is not provided in the JSON file.

```ts
{
sumAuto:true,
sum: undefined
}
```

test-e2e/panels/spectra.test.ts
hamed-musallam: I took the value directly from the color function https://github.com/cheminfo/nmrium/blob/84ea3e4ef8471d60a8bec508410771e3cd0331c1/src/data/utilities/getColor.ts#L2

test-e2e/panels/ranges.test.ts
hamed-musallam: @lpatiny 

what do you think we have to do in this case, should be all the options on the side nmr-load-save, or should be NMRium which I think it is better because it directly related to NMRium view

test-e2e/panels/spectra.test.ts
hamed-musallam: @lpatiny 

Can I know why it is not good to use it like that from your point of view?

test-e2e/panels/ranges.test.ts
hamed-musallam: we need to add comments in the test to clarify this behavior

targos:
We reverted https://github.com/cheminfo/nmrium/commit/bbdad09b634f0c73efc8f9851e204a635fb62e4a.
IIUC, https://github.com/cheminfo/nmrium/pull/1816 is going to handle loading from inline JSON.

PR #1821, opened 2022-10-21
refactor: contours rendering

hamed-musallam:
Old State
Contours calculation is done when initiating the ft 2d spectrum or fire Wheel event and saving the generated data inside the contours object per spectrum.

``` ts
{
data: [
          // spectra list
            {
             processing : <processingObject>
            }
          ]
          // contains generated contours per spectrum
contours: {}
}
```
New State

We removed all the previous objects in state and replace them with a new object under ```view``` object which holds the current level of the contours per spectrum.

The Contours calculation is done in the view component

``` ts
{
  view:{
           zoom:{ 
                     levels: {[spectrum id]: {positive:number,negative:number} }
                     }
          }
}
```


src/component/loader/DropZone.tsx
```diff
@@ -58,9 +58,8 @@ function DropZone(props) {
         .then((fileCollection) => {
           dispatch({ type: LOAD_DROP_FILES, fileCollection });
         })
-        .catch((e: any) => {
+        .catch(() => {
           dispatch({ type: SET_LOADING_FLAG, isLoading: false });
-          reportError(e);
```
targos, line 63: Why did you remove this?

hamed-musallam: Because we are not handle it at the level of nmrium, unless if the purpose of it to let others to listen for the error event outside nmrium.

targos: I added it for us developers. The default behavior is similar to `console.error(e)`

hamed-musallam: ok, I will revert the changes

hamed-musallam:
@targos 

can you check this PR

src/component/2d/Contours.tsx
```diff
@@ -1,115 +1,169 @@
...
-            );
-            pathBuilder.lineTo(
-              _scaleX(lines[i + 2] + xShift),
-              _scaleY(lines[i + 3] + yShift),
-            );
-          }
-        }
+  const level = useContoursLevel(datum, sign);
+
+  useLayoutEffect(() => {
```
targos, line 104: Why do you use a layout effect to set the opacity?

hamed-musallam: No for a specific reason and it does not matter if I use useLayoutEffect or useEffect in this case.

targos: My question is why do you use an effect at all

hamed-musallam: In the previous version of my code, I use mem hook for the whole path component which was wrong because in any way the component will need to be rendered if we zoom in/out or change the level of the contours so then I memorized the contours calculation to prevent the contours re-calculation when zooming in and out and forgeted to remove the ref and the useEffect that i use for change the color of the contours.

 Thank you for spotting the light on this part of the code, I will remove it and fix the code

PR #1860, opened 2022-11-01
feat: database table scroll position

### LINKED ISSUE #1854: Database panel and display scroll position
When scrolling in the database panel would it be easily possible to display the current position and the total number of rows ?

hamed-musallam:
a demo for a sticky footer counter under the table when the virtual scroll activated 

close #1854


lpatiny:
Looks good thanks. Could you add a small transition of 1s to dim it out like in the animated gif of the issue ? After you may merge it.

Issue #1763, opened 2022-09-21
Add test case on processed spectra


lpatiny:
When we process spectra in NMRium we can save the result as a JSON. In this json we also have all the 'filters' that are applied in the spectrum.

It is easy to break something and we should test as much as possible that the result of loading this 'nmrium' file is consistent.

@wadjih-bencheikh18 Please could you add a testcase:

That loads

![image](https://user-images.githubusercontent.com/1484241/191554276-a757b5b5-f3a6-444f-923d-d2c26632694a.png)

That checks first and last line of the peaks (value in ppm) and the number of lines

![image](https://user-images.githubusercontent.com/1484241/191554377-fb7cc220-c745-4158-afd6-a44ba5e474f8.png)

Check that we have the right number of filters

![image](https://user-images.githubusercontent.com/1484241/191554434-4ae292ea-52ce-4fec-ac79-d9b22996e9ae.png)

If possible try to find a way to check that there is no negative peaks (just tell me if it is too hacky or not feasible).

![image](https://user-images.githubusercontent.com/1484241/191554628-a6191e54-1d70-40f0-b4b6-4053c0ff5bc4.png)

A problem that could occur is that we break the code and we obtain something like:

![image](https://user-images.githubusercontent.com/1484241/191554779-b5dc5371-0dec-4d47-97c2-a9ea35b8b412.png)

PR #1864, opened 2022-11-03
test: processed spectra

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/1763


test-e2e/panels/filters.test.ts
```diff
@@ -1,165 +1,183 @@
-import { test, expect } from '@playwright/test';
```
lpatiny, line 1: Could you checked what happened with this commit ?

The end of line is expected to by LF (ascii 10) and in this commit it is now CR + LF (ascii 13 + ascii 10)

wadjih-bencheikh18: fixed

PR #1887, opened 2022-11-08
feat: order multiple spectra analysis table

### LINKED ISSUE #1870: Reorder legend of multiple spectra
According the intensity we should have an option in the multiple spectra preferences panel to sort by increasing / decreasing value.

### LINKED ISSUE #1876: Reorder spectra and columns in multi spectra analysis
We should have a way to click in the header of the columns in the multiple spectra analysis result

image

And we should not only reorder the column but also the spectra.

There should be 3 status: increasing, decreasing, original order.

hamed-musallam:
close  #1876
close  #1870


lpatiny:
Please take care that it is sorted as numbers and not string

image

lpatiny (approved):
Works as expected. Many thanks !

PR #1918, opened 2022-11-15
feat: show molecule atoms number

### LINKED ISSUE #1916: Add toggle button to display atom number
In order to reload the assignment from other sources we need to be able to debug and display the atom number.

Please add a toggle button in the structure panel:

image

That will add the property `showAtomNumber` in OCLnmr component.
```
            <OCLnmr
              showAtomNumber
```

hamed-musallam:
refactor: molecule view object, use an object instead of an array, and change the object structure

close #1916


lpatiny (approved):
It works as expected thanks !

PR #1919, opened 2022-11-16
refactor: spectra list

### LINKED ISSUE #1881: Add header in the list of spectra
image

We should display the column name in a header.

### LINKED ISSUE #1884: Spectra list and meta / info information
From preferences we should be able to add unlimited number of columns by selecting jpath.

This is related to:

https://github.com/zakodium-oss/analysis-ui-components/issues/252

### LINKED ISSUE #1928: Display signal not ranges if no couplings
Currently if the multiplicity is 'm' or empty we display the range (from - to) rather that the signal (delta) in the first column.

image

For 'm' this should stay like that but in case nothing is defined and there is a signal with delta please display the delta and not the range.

### LINKED ISSUE #1934: Add toggle button to show / hide peaks
At the level of the peaks panel we should add a toggle button to show / hide the peaks (rather than at the level of the spectrum list)

### LINKED ISSUE #1935: Spectra tabs and superscripts
Would be nice to format isotopes as isotopes

Using a code similar to:

`'1H,13C'.replace(/(\d+)(?=[A-Z])/g,'<sup>$1</sup>')`

image

### LINKED ISSUE #1940: Integration not updated when changing spectrum
While the integral is correctly updated when we apply new filters the integration (relative and absolute) is not updated.

This should be the case in the 'integral' tool and 'range' tool.

hamed-musallam:
close  #1881
close  #1884
close  #1934
close  #1935
close  #1940
close  #1928


hamed-musallam:
@hamed-musallam 

basic table rows sort by drag and drop, and still needs more improvement before I can use it 

https://codesandbox.io/s/drag-and-drop-table-p149g9?file=/src/App.js

lpatiny:
Functionalities looks ok, thanks !

Please remove 'peaks' from this table. Just add a toggle button at the level of the peaks panel (this can be another issue) : https://github.com/cheminfo/nmrium/issues/1934

Could you add the '+' at the level of each line so that we can add a line anywhere in the table ?

lpatiny:
We currently have couple of problems: 

@jobo322
- the test cases are failing
- the `13C spectrum` example does not load
- the auto range picking in nmr-load-save does not lead to the same result as the one in NMRium

@hamed-musallam 
- when resurrecting spectra the integration is too big
- integration is not calculated on the currently displayed spectrum (if we remove a filter the integration (relative and absolute) should be recalculated (both in ranges and integrals) https://github.com/cheminfo/nmrium/issues/1940

PR #1952, opened 2022-11-25
chore: update nmr-load-save to 0.6.0

jobo322:


targos (changes requested):
https://github.com/cheminfo/nmr-load-save/issues/67 should be fixed and released first

PR #1965, opened 2022-11-29
feat: nmr-load-save prerelease with filters

jobo322:


lpatiny:
With those settings (with experiment 1 that does not exists)

image
 
There should be no spectra loaded but still there are some.

PR #1966, opened 2022-11-30
feat: improve general settings UI

hamed-musallam:
- replace the reset button title with an icon
- disabled reset button if
 1- not predefined workspace
 2- no changes

- add the version number to workspaces dropdown list
- copy workspace to clipboard
- set default height of the preferences dialog box could be 50% of NMRium component size


lpatiny:
There is an issue if you load the example:

http://127.0.0.1:3000/#/SamplesDashboard/ztr2yvyl9z/CustomWorkspacedataxtcdataxtcXTCjson

And try to change the workspace from the general preferences you get the following error:

image

hamed-musallam:
@lpatiny 

It is not practical to hide them behind the 'experimental features' flag, First you have to change flag value to true in every workspace to use them and if you set flage to true and move to other workspace where flag value is false the buttons will not be visible.

hamed-musallam:
@lpatiny 
'experimental features' flag becomes affected once you click `save` button and not when change the value.

hamed-musallam:
@lpatiny 

Can you remove your local storage and test again?

PR #1969, opened 2022-12-01
fix: baseline correction live preview

### LINKED ISSUE #1941: Baseline correction of degree 3 does not work.
Polynomial baseline correction of degree 3 seems not to work while levels 2 or 4 are ok.

@jobo322 @hamed-musallam Not sure where the bug is.

### LINKED ISSUE #1949: Baseline correction preview is not working
Even if preview is activated we currently don't see the baseline preview.

hamed-musallam:
close  #1949
close  #1941


lpatiny (approved):
Works perfectly now.

PR #1974, opened 2022-12-01
chore: improve metabo workspace

lpatiny:


hamed-musallam:
@lpatiny 

you do not need to list the tools or panels which is hidden because by default we assume it is hidden unless you set them as visible

src/demo/views/CustomWorkspace.tsx
```diff
@@ -12,38 +12,39 @@ const customWorkspaces: CustomWorkspaces = {
...
       },
       toolBarButtons: {
         baselineCorrection: false,
-        exclusionZones: false,
-        exportAs: false,
+        exclusionZones: true,
+        exportAs: true,
         fastFourierTransform: false,
         import: true,
         integral: false,
```
hamed-musallam, line 35: you can remove it because it is by default hidden and the same for the panels where it the `display` value is false

lpatiny: In fact I just copied the preferences from the config panel and pasted them without any changes. You think we should not copy the values that are false ? This could be a problem if the default value changes in the future.

hamed-musallam: you can keep them like that or remove the default ones, in any way, I have a file for the default workspace values where I merge it with the values that come from you or the predefined workspace to be sure that all the objects are listed there.

PR #1975, opened 2022-12-02
fix: edit range

### LINKED ISSUE #1945: Bug is not specifying coupling constant in ranges
If we edit a range and change a multiplet to doublet for example

Without specifying the coupling constant and we reedit it we get a crash:

image

hamed-musallam:
clone the range once we open the editing modal and use the clone range to rest the changes once close and not save the changes

close #1945


lpatiny:
This PR still does not fix the bug seems to me.

hamed-musallam:
@lpatiny 

i fixed the form validation, it should work now

PR #1977, opened 2022-12-02
refactor: nmrium object from nmr-load-save

jobo322:


src/component/loader/DropZone.tsx
```diff
@@ -62,7 +62,7 @@ function DropZone(props) {
       const fileCollection = await fileCollectionFromFileList(files);
 
       const { nmrLoaders: filter } = preferences.current;
-      const data = await readDropFiles(fileCollection, { filter });
+      const { data } = await readDropFiles(fileCollection, { filter });
```
lpatiny, line 65: I guess you should at least check for the append flag (true / false) to know if the full data state should be replaced.

jobo322: @lpatiny, I did ask about it but @hamed-musallam will make it directly in the PR

src/component/reducer/actions/LoadActions.ts
```diff
@@ -97,80 +116,44 @@ function setPreferences(draft: Draft<State>, data) {
...
-      draft.data.push(initiateDatum1D(spectrum, draft.usedColors));
-    } else if (info.dimension === 2) {
-      draft.data.push(initiateDatum2D(spectrum, draft.usedColors));
-    }
-  }
-  for (let molecule of molecules) {
-    MoleculeManager.addMolfile(draft.molecules, molecule.molfile);
+function loadDropFiles(draft: Draft<State>, action) {
+  const { append } = action.payload;
+  if (!append) {
```
lpatiny, line 148: Please be positive ;) 

And invert the condition (and the 2 blocs)

PR #2003, opened 2022-12-08
refactor: rework the preferences

hamed-musallam:


lpatiny:
In the dialog please use 'Cancel' and 'Save' for the button names.

image

When opening this dialog asking for a workspace name take care that the cursor focus goes directly on the text input field.

For now when I save as a new name the default workspace I have no more options:

image

I would use a save icon rather than the work 'Save' with mouse hover explaining 'Save workspace in locally in the browser'.

The full process works correctly when you start from a Custom Workspace.

I'm still not convinced by the buttons in the general settings

image

Should we have the same 'Save' icon next to the drop down menu of the workspace and as button 'Cancel' / 'Apply' ? It would be the same logic as in the preferences of a panel. You can just apply / cancel but it is not saved.

PR #2010, opened 2022-12-15
fix: dropdown list position

hamed-musallam:


src/component/elements/dropDownButton/DropDownList.tsx
```diff
@@ -58,16 +58,20 @@ function DropDownList({
   labelKey,
 }: InnerDropDownListProps<DropDownListItem>) {
   const ref = useRef<HTMLDivElement>(null);
+  const containerBoundariesRef = useRef<DOMRect>();
```
lpatiny, line 61: Why are you caching those values because it is never reused ?

hamed-musallam: in the development mode because of the strict mode it's triggered twice

lpatiny: Because otherwise it does not work in development mode otherwise ?

lpatiny: There is specific information in the documentation if there is a problem:

https://beta.reactjs.org/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development

You may have to add a cleanup callback.

hamed-musallam: > Because otherwise it does not work in development mode otherwise ?

yes exactly

hamed-musallam: i think we do not need a clean-up callback because the next getBoundingClientRect() will take into account the `left` value that we add

targos: Your cleanup callback should undo what you do in your effect. In this case, reset the value of `ref.current.style.left`. Then when it runs a second time, it will work correctly.

hamed-musallam: Yes, you are right, I was thinking instead of resetting the value, I can cache it.

PR #2023, opened 2022-12-19
feat: improve matrix generation

hamed-musallam:


src/data/data1d/filter1d/signalProcessing.ts
```diff
@@ -0,0 +1,83 @@
+import * as Filters from 'ml-signal-processing/src/filters/filters';
```
targos, line 1: @hamed-musallam This is your bug. you shouldn't be deep-importing from this package

hamed-musallam: Thanks !!! @targos

PR #2036, opened 2023-01-02
feat: step by step user guide 

hamed-musallam:


hamed-musallam:
@targos 

the `react-joyride` raise an error because it use  `global`  object, `global is not defined`, https://github.com/gilbarbara/react-joyride/issues/772.

This problem will not be resolved soon from them so what we can do?

hamed-musallam:
@targos 

the only way that i found to use this package is by pass array of elements to `steps` property, can you review the code.

Issue #2044, opened 2023-01-05
Tests of Jcamp upload


lpatiny:
Currently we are testing the upload of .nmrium file by simulating a click and open file.

However we are not testing the upload of a JCAMP-DX file and there is currently a bug related to this feature: https://github.com/cheminfo/nmrium/issues/2042

This test is similar to:

https://github.com/cheminfo/nmrium/blob/main/test-e2e/core/import.test.ts

You may create already this test and 'expect to fail'.

PR #2055, opened 2023-01-08
test: drag and drop JCAMP-DX file

wadjih-bencheikh18:
closes https://github.com/cheminfo/nmrium/issues/2044


test-e2e/core/import.test.ts
```diff
@@ -165,3 +165,9 @@ test('should load multiple files using drag and drop', async ({ page }) => {
   // If the file was loaded successfully, there should be many tabs.
   await expect(nmrium.page.locator('_react=Tab[tabid = "13C"]')).toBeVisible();
 });
+test.only('should load JCAMP-DX file', async ({ page }) => {
```
targos, line 168:
```suggestion
test('should load JCAMP-DX file', async ({ page }) => {

```

targos:
@lpatiny Is it expected that the test passes on Chromium right now?

PR #2060, opened 2023-01-10
refactor: enable virtual scroll on the spectra list

### LINKED ISSUE #1981: Slow highlight in list of spectra panel
There seems to be a general problem with react-table component that makes it very slow. I wonder if this can be solved globally.

When we have a lot of spectra (1000 for example) and even if the spectra are not displayed it takes around 1s to refresh the list of spectra ?!

image

and it seems to be related to react-table

image

hamed-musallam:
close #1981

### DIFF
--- a/src/component/panels/SpectrumsPanel/SpectraTable.tsx
+++ b/src/component/panels/SpectrumsPanel/SpectraTable.tsx
@@ -36,6 +36,14 @@ export const SpectraTableButtonStyle: CSSProperties = {
   margin: 'auto',
 };
 
+const jPathColumnStyle: CSSProperties = {
+  width: '10%',
+  maxWidth: 0,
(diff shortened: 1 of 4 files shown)


test-e2e/type/ClickOptions.ts
```diff
@@ -0,0 +1,17 @@
+interface ClickOptions {
```
targos, line 1: You should be able to use the types from playwright.

hamed-musallam: it seems that they did not expose it, this why I create an interface

targos: `type ClickOptions = Parameters<Page['click']>[1];`

hamed-musallam: Thanks  @targos

PR #2061, opened 2023-01-11
refactor: multiple spectra analysis cells and the highlight

### LINKED ISSUE #1913: Multiple spectra analysis and highlight
When the mouse is over a column in multiple spectra analysis we should highlight the column and the corresponding zone in the spectra.

image

hamed-musallam:
close #1913


lpatiny:
Seems the sorting has an issue. It looks like it sorts and then refresh quickly to another value but the resulting sort is not correct.

PR #2068, opened 2023-01-12
feat: spectra info block

hamed-musallam:


lpatiny:
The block info looks perfect thanks !

What about the line customization of the tracker ? Did you keep the code you did before  somewhere ?

image

I guess it should go in the preferences of the multiple spectra analysis

image

But can be part of another PR if you prefer.

hamed-musallam:
@lpatiny 
Yes, i prefer to have it in another PR because I need to make some refactoring there before i add the feature

hamed-musallam:
@lpatiny 

i append the legends customization  to this PR

lpatiny:
I tried to customise the coffee but got an error

image


image

lpatiny:
Also when you display the legend and you are 'outside' the spectrum we should not remove '00000' everywhere. Just remove completely the information

image

PR #2087, opened 2023-01-24
refactor: disable 'export as data source' if the 'baseURL' is not exists

### LINKED ISSUE #2084: disable export as data-source if the spectra do not have baseURL

hamed-musallam:
close #2084


lpatiny:
If we are saving the .nmrium file without data we still obtain a source.



```json
          "source": {
          "files": [
            {
              "name": "1h.jdx",
              "size": 175555,
              "relativePath": "1h.jdx",
              "lastModified": 1614490027000
            }
          ],
          "filter": {
            "jcamp": {
              "index": 0
            }
          }
        },
```

I think that if we don't want toe `data`, we should not have the `data` property in the .nmrium file at all.

```
{
  "version": 4,
  "data": {
    "spectra": [
      {
```

PR #2090, opened 2023-01-24
test: fix auto assignments test

hamed-musallam:


hamed-musallam:
@targos 

can you give me a hand with this:

I changed the automatic assignment button state from disable to enable for the signs that the ranges and zones are ready, but it seems not to work well, Do you have any suggestions?

targos:
@hamed-musallam 

> I changed the automatic assignment button state from disable to enable for the signs that the ranges and zones are ready, but it seems not to work well, Do you have any suggestions?

I think this is not a reliable way to do it. The reason is that you may miss the state during which the button is disabled.
The best way to deal with these things is to wait for some effect of the button that will stay indefinitely before the next action.

In the case of auto assignments, if I understand correctly, the click on the first button (automatic ranges) will add some labels in the spectrum:
image

I would add an `await expect(page.locator('some locator for one of the labels')).toBeVisible()` or `await expect(page.locator('some locator for all the labels')).toHaveCount(expectedCount)`

Then the click on the second button (automatic assignment) will create a table:
image

Here you can expect that some content in the table have some valu. For example here that the score is 0.75.

hamed-musallam:
@targos 

> I would add an `await expect(page.locator('some locator for one of the labels')).toBeVisible()` or `await expect(page.locator('some locator for all the labels')).toHaveCount(expectedCount)`

we already waiting for the ranges notations to be visible, but this does not resolve the issue and the big question for me is why it works locally !!!

https://github.com/cheminfo/nmrium/blob/de448bf27528bfdadf7d2bdbb429a8ed042ce536/test-e2e/panels/autoassignment.test.ts#L37-L39

hamed-musallam:
@targos 

> I think this is not a reliable way to do it. The reason is that you may miss the state during which the button is disabled.
The best way to deal with these things is to wait for some effect of the button that will stay indefinitely before the next action.

the playwright auto-waiting the actions, and in our case, it will wait for the button to be clickable otherwise the action fails with the TimeoutError.

https://playwright.dev/docs/actionability

targos:
I added some configuration to upload test traces when there are failures. TBH I don't understand how you're debugging this without something like that in place!

hamed-musallam:
@targos 

I did not know that we can trace the playwright test result, I always check the playwright test failure result from the GitHub action

hamed-musallam:
@targos 

I can not see the generated trace file, where i can find him?

targos:
> I did not know that we can trace the playwright test result, I always check the playwright test failure result from the GitHub action

I did not know either until today. I searched in the Playwright documentation how to debug errors in CI.

> I can not see the generated trace file, where i can find him?

From the workflow run summary: https://github.com/cheminfo/nmrium/actions/runs/4043243064

hamed-musallam:
@hamed-musallam 

as I can see the file should be in this path  "show-trace test-results/panels-autoassignment-automatic-assignment-panel-firefox/trace.zip" but it is not

targos:
it is: 
image

PR #2091, opened 2023-01-25
refactor: floatting molecule structures

hamed-musallam:


lpatiny:
I would prefer we see the square being resized because currently I don't have visual feedback when I receive the molecule.

hamed-musallam:
> I would prefer we see the square being resized because currently I don't have visual feedback when I receive the molecule.
> 
>  [  ](https://user-images.githubusercontent.com/1484241/214881901-9a5d0e42-b77d-4a78-8ae6-e7aa3a1fa075.gif) [ ](https://user-images.githubusercontent.com/1484241/214881901-9a5d0e42-b77d-4a78-8ae6-e7aa3a1fa075.gif)

Done, can you check it one more time

src/component/NMRium.tsx
```diff
@@ -360,13 +361,16 @@ function InnerNMRium({
...
+                                      id="nmrium-viewer"
                                       data-test-id="viewer"
                                       ref={viewerRef}
                                       style={{
                                         width: '100%',
                                         height: '100%',
+                                        position: 'relative',
                                       }}
                                     >
+                                      <FloatMoleculeStructures />
```
targos, line 373: It seems you wanted to move the rendering of FloatMoleculeStructures to this place, but it is still rendered a second time elsewhere in the DOM (you can search for it in the react dev tools, the other render has a size of 0x0)

src/component/1d-2d/components/FloatMoleculeStructures/ActionsButton.tsx
```diff
@@ -5,7 +5,7 @@ import Button from '../../../elements/Button';
 
 export default function ActionsButton({ onFloatBtnClick }) {
   return (
-    <div data-no-export="true" style={{ display: 'flex', zIndex: 0 }}>
+    <div id="float-molecule-actions" style={{ display: 'flex', zIndex: 0 }}>
```
targos, line 8: It is not allowed to have multiple elements with the same id in html (this can happen when more than one molecule is floated)

src/component/NMRium.tsx
lpatiny: @hamed-musallam Could you explain why you moved it ? Because moving the molecule was already working correctly so we expected that resizing would work as well.

src/component/1d-2d/components/FloatMoleculeStructures/ActionsButton.tsx
hamed-musallam: Yes, you are right, I will change it to class.

lpatiny:
If we have 2 molecules and we try to save there is a crash

image

image

src/component/1d-2d/components/FloatMoleculeStructures/ActionsButton.tsx
targos: You resolved it but it's still an id (and not unique).

targos: I'd keep an id, but include the uuid of the molecule in it to make it unique.

src/component/utility/export.ts
```diff
@@ -268,6 +268,37 @@ function getBlob(rootRef: HTMLDivElement, elementID: string): BlobObject {
...
+    `#nmrium-viewer`,
+  );
+
+  const floatingMoleculesGroup = document.createElement('g');
+
+  for (const element of nmriumViewer.querySelectorAll('.draggable-molecule')) {
+    const transform = window
+      .getComputedStyle(element)
+      .getPropertyValue('transform');
+    const matrix = new WebKitCSSMatrix(transform);
```
targos, line 282: it seems that the standard name for this is `DOMMatrix`: https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix

hamed-musallam: the WebKitCSSMatrix is an alias to DOMMatrix

src/component/1d-2d/components/FloatMoleculeStructures/ActionsButton.tsx
hamed-musallam: Sorry, it is a mistake I did not commit the changes, I need to use class instead of id because i need still be able to access this element to get the height

hamed-musallam: https://github.com/cheminfo/nmrium/pull/2091/commits/6eb32513f65b6ebabb7f52e27caa4f4a22ff905b

src/component/utility/export.ts
lpatiny: Yes this is the whole point. To use the preferred name and not an alias.

hamed-musallam:
> If we have 2 molecules and we try to save there is a crash
> 
> image
> 
> image

i fixed it, you should not face this problem anymore

PR #2092, opened 2023-01-27
refactor: active spectrum

hamed-musallam:


src/component/reducer/helper/getSpectrum.ts
```diff
@@ -7,8 +7,10 @@ export function getSpectrum(state: Draft<State>, index: number);
 export function getSpectrum(state: Draft<State>, id: string);
 
 export function getSpectrum(state: Draft<State>, value?: number | string) {
-  const activeSpectrum =
-    state.view.spectra.activeSpectra[state.view.spectra.activeTab];
+  const { activeSpectra, activeTab } = state.view.spectra;
+
+  const activeSpectrum = activeSpectra[activeTab]?.[0];
```
targos, line 12: I think this function should only return a spectrum if the active spectra array contains exactly one element.

hamed-musallam: Yes you are right, this was a draft PR, just i change it too fast to see what I have to change and where this should affect

PR #2094, opened 2023-01-27
new structure of nmrium file and saver from nmr-load-save

jobo322:


lpatiny:
This error message is really weird because it speaks from dataURL while it is just a file without data which is allowed.

[test (13).nmrium.zip](https://github.com/cheminfo/nmrium/files/10532973/test.13.nmrium.zip)

lpatiny:
I should also be able to drag / drop a .nmrium file that do not contain `data` in an instance that has already some spectra and the 'settings' for example should change.

lpatiny:
The current version is also buggy. If you try to save the data source of one of the example (like ethylbenzene) you get an error:

image

Issue #1634, opened 2022-07-21
Jres should have his own tab


lpatiny:
It is quite annoying to superimpose COSY and Jresolv.

How we could deal with this to have a different tab ? Any idea @jobo322 

![image](https://user-images.githubusercontent.com/1484241/180173725-1bd2c9bb-7729-4905-bb91-ad035e6faf8f.png)

jobo322:
yes, we should have a tab for DOSY or JRES because the indirect dimension is not in ppm. I think those spectra should not be superimposed with any other.

lpatiny:
We could have as second nucleus 'J' so there will be a tab '1H,J'

jobo322:
it is related with this ticket

-https://github.com/cheminfo/nmr-load-save/issues/54

@lpatiny with this I think it will generate a own tab.

jobo322:
it was done by fb135a7b7c53df58317f6217609bf2319f483f05

PR #2106, opened 2023-02-01
new LDR supported and phase correction if parameters exists

### LINKED ISSUE #1152: JResolv example
Do we have a testcase of Jresolv on dev.nmrium.org ?

jobo322:
.SHIFT REFERENCE is used to calculate the frequencyOffset that is used to calculate the first point value of ppm axis. it is ignore if SFO1, BF1 exists.

REVERSE is checked after FFT filter is applied to reverse the data.

close #1152 #1634

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "ml-stat": "^1.3.3",
         "multiplet-analysis": "^2.0.0",
         "nmr-correlation": "^2.3.3",
-        "nmr-load-save": "^0.7.17",
+        "nmr-load-save": "0.7.17-pre.1675262677",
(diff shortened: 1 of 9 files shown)


hamed-musallam:
> @hamed-musallam I did add the phaseCorrection after add the FFT filter in the same place, are you agree with it?

No, I think we should not have it like that, Why do you do it like that?

jobo322:
> > @hamed-musallam I did add the phaseCorrection after add the FFT filter in the same place, are you agree with it?
> 
> No, I think we should not have it like that, Why do you do it like that?

The Idea is to apply the phaseCorrection after FFT if the parameters exists.

PR #2107, opened 2023-02-02
refactor: spectra and auto assignments panel bar buttons

hamed-musallam:
- [x] refactor the spectra and auto assignments panel bar buttons
- [x] delete/show/hide spectra based on the selected spectra
- [x] select all the spectra if the mouse over the spectra table and press Ctrl + a 
- [x] prevent show spectra when select spectrum/spectra are hidden  
- [x] Fix tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "react-error-boundary": "^3.1.4",
         "react-icons": "^4.7.1",
         "react-inspector": "^6.0.1",
+        "react-kbs": "^2.1.0",
         "react-mf": "^2.0.2",
         "react-ocl": "^5.0.0",
         "react-ocl-nmr": "^2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,6 @@
         "react-error-boundary": "^3.1.4",
         "react-icons": "^4.7.1",
         "react-inspector": "^6.0.1",
-        "react-kbs": "^2.1.0",
         "react-mf": "^2.0.2",
         "react-ocl": "^5.0.0",
         "react-ocl-nmr": "^2.0.0",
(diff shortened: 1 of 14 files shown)


lpatiny:
Rather that stealing the focus (that leads to bad side effect like if you move your mouse when editing a peak chemical shift) it would be better to assign globally 

CTRL (CMD on mac) + A to select all spectra. This can even work if the spectra panel is not open.

hamed-musallam:
> Rather that stealing the focus (that leads to bad side effect like if you move your mouse when editing a peak chemical shift) it would be better to assign globally
> 
> CTRL (CMD on mac) + A to select all spectra. This can even work if the spectra panel is not open.

The problem we use the same shortcut for other things, especially for Phase correction, and Apodization

lpatiny:
For phase correction and apodization we don't use any modifiers key.

So CTRL (CMD) A is currently not use and it selects all the text in the browser by default.

hamed-musallam:
the side effect is acceptable because in any way each time you enter over the table, we get the focus and vice versa

So if we take your example, when we edit the  "peak chemical shift" we lost focus, this is normal behavior and I think is acceptable because it does not make sense to have a "select all" option and I'm not on the spectra table

hamed-musallam:
@lpatiny 

the apodization and phase correction triggered because we only check if !e.shiftKey && !e.metaKey

hamed-musallam:
@lpatiny 

As i mentioned, if we use this event globally then the only way to know whether I'm on the spectra list or not is by looking inside the objects list on which listeners will be invoked `composedPath`.

 from my point of view, it does not make sense to trigger the select all and the mouse is not over the spectra table

lpatiny:
On my computer CMD + A select the page so I don't understand your answer.



Currently taking the focus leads to bugs.

lpatiny:
It should work where ever is the mouse so in fact it simplifies the code.

hamed-musallam:
@lpatiny 

it triggers the phase correction for me 
image

hamed-musallam:
@lpatiny 

> Currently taking the focus leads to bugs.

Which kind of problem do we have when wee lost focus?

PR #2135, opened 2023-02-09
refactor: move 'verticalAlign' object to 'view' object

hamed-musallam:
```[tasklist]
- [X] move `verticalAlign` object to `view` object
- [X] refactor the vertical-align object and calculate the vertical shift on fly
- [ ] refactor samples
```


src/component/reducer/Reducer.ts
```diff
@@ -122,6 +123,11 @@ export interface ViewState {
   zoom: {
     levels: ContoursLevels;
   };
+  /**
+   * options to control spectra vertical alignment
+   * @default  'bottom'
+   */
+  verticalAlign: Partial<Record<Nuclei, VerticalAlignment>>;
```
targos, line 130: Shouldn't we have a more "generic" Record at this level with an object of properties? I expect more values to depend on the nucleus and having a separate record for each of them seems complicated.

hamed-musallam: Actually, we depend on that in different places, 

```ts
  spectra: {
    /**
     * active spectrum id per nucleus
     * @default {}
     */
    activeSpectra: Record<Nucleus , ActiveSpectrum[] | null>;
    },
verticalAlign: Partial<Record<Nuclei, VerticalAlignment>>
```
 but I think that it is more precise and specific to have it like that. for example, verticalAlign is only for 1d so it does not make sense to have this value  for a 2d nucleus 

```ts
 { "1H,1H" :{
     verticalAlign,
     activeSpectra
     }
   }
```

hamed-musallam: @targos 

What do you think ?

targos: Then what about an object for all 1D-specific parameters which depend on the nucleus?

hamed-musallam: @targos 

Yes, it could be like that, let's assume we have 

```ts
view:{
 1H: {
   verticalAlign,
 activeSpectrum,
    ...etc
  },

},

```
 in above structure how I can force specific type for the object based on the Nucleus ?


## Or

```ts
view1D:{
    1H: {
       verticalAlign,
     activeSpectrum,
        ...etc
      },
}

view2D:{
  "1H,1H":{
     activeSpectrum,
     ...etc
    }
}

```



on the other hand, we will consider that all object is not optional so we should initiate the object with an initial value, which could be `null` .  This happens on the first time, we attend to store a value, for example, the active spectrum and the object `1H` object does not exist.

```ts
{
  1H: { 
         verticalAlign: "bottom",
          activeSpectrum:  1,
          x:null,
          y: null,
         ...etc
          }
}
```

targos: I think we should avoid mixing 1D and 2D when they have different interfaces, so my preference goes towards `view1D`/`view2D`.

hamed-musallam: Ok, then we should have 3 objects

```ts

view: {
// which contains all objects that are grouped by the nucleus.
 },
view1D,
view2D

```

What do think about the other object that is per spectrum, for example, 

```ts

{
  ranges: Array<RangeToolState>;
  zones: Array<ZoneToolState>;
  /**
   * peaks view property
   * where the key is the id of the spectrum
   */
  peaks: Record<string, PeaksViewState>;   
  zoom: {
    levels: ContoursLevels; // for 2D only
  }
}

Comment (hamed-musallam) [reply src/component/reducer/Reducer.ts:130]: What do you think about this structure ?

```ts 
view: {
  nuclei:{
    "1D":{
      "1H":{
         verticalAlign: "bottom",
         activeSpectra:"",
         selectReferences:null,
        }
     },
     "2D":{
     "1H,1H":{
         activeSpectra:"",
         selectReferences:null,
        }
       }
   },
  spectra: {
    /**
     * current select tab (nucleus)
     * @default null
     */
    activeTab: string;

    /**
     * show the spectra legend and the intensity
     * @default false
     */
    showLegend: boolean;
  }
  }
```

targos: > Ok, then we should have 3 objects

No, we should have everything inside of view:
```
view: {
  nuclei1D: { // Record<string, ViewNuclei1D>
    '1H': {},
    '13C': {},
  },
  nuclei2D: { // Record<string, ViewNuclei2D>
    '1H,1H': {},
  },
  spectra1D: { // Record<string, ViewSpectra1D>
    uuid1: {},
    uuid2: {},
  },
  spectra2D: { // Record<string, ViewSpectra2D>
    uuid3: {},
    uuid4: {},
  },
  prop1, // other settings that don't depend on nucleus or spectrum
  prop2,
}
```

hamed-musallam: @targos  thanks, we need to consider this refactoring before having a new release of NMRium otherwise we have to write a migration script

PR #2126, opened 2023-02-07
feat: add the 'source' property for original file

jobo322:
fix: fix parse ACS string (resurrect)
closes: https://github.com/cheminfo/nmrium/issues/1251

feat: add nmr-load-save source
closes: https://github.com/cheminfo/nmrium/issues/2148

close: #2146

PR #2136, opened 2023-02-09
fix: load bruker folder without expno

jobo322:


jobo322:
the current state of nmrium is not compatible with nmr-load-save 0.7.22 so this work will be merged with the existing PR 

- https://github.com/cheminfo/nmrium/pull/2126

to avoid re-do the fixing for processed13C sample test

Issue #2175, opened 2023-02-22
Multispectra spectra analysis testcase


lpatiny:
We need to create a new e2e test case called: 'Multiple spectra analysis'

This will load the coffee data set

1. check that we have 13 spectra
2. 2. Check the spectrum names: coffee 1 till coffee 13
3. check that the color of the spectra are different (13 different colors)
4. right click on Spectrum Name and recolor : spectra should keep different color (13 different colors)
5. click on pulse and 'recolor', should have all the same color
6. click again on Spectrum Name and recolor: should have 13 colors
7. click on Solvent and recolor: should all have the same color
8. click in the header for recolor ![image](https://user-images.githubusercontent.com/1484241/220538633-7afb235f-704d-4e53-b461-253fed4a95ae.png): should have all different colors

You should add a helper like 'getNumberOfDistinctColors' for an array of spectra probably.

https://1251-resurrect-spectrum.nmrium.pages.dev/#/SamplesDashboard/0nyxdgi5qvq/View

PR #2189, opened 2023-02-23
test: Multiple spectra analysis

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/2175


wadjih-bencheikh18:
@lpatiny there's an error in `Check again recolour based on spectrum name` test

PR #2195, opened 2023-02-23
refactor: onChange call-back

hamed-musallam:


src/component/NMRium.tsx
```diff
@@ -126,10 +127,20 @@ export type NMRiumWorkspace =
...
+  data: SerializedNmriumState,
+  source: 'data',
+) => void;
+export type OnViewChange = (data: ViewState, source: 'view') => void;
+export type OnSettingsChange = (
+  data: WorkspacesWithSource,
+  source: 'settings',
+) => void;
+export type OnNMRiumChange = OnDataChange | OnViewChange | OnSettingsChange;
+
```
lpatiny, line 140: The callback should always receive the same object, the NmriumState. It does not matter if it is a data change, view change or settings change, it is always a NmriumState.

Related topic all this code should likely disappear:

https://github.com/cheminfo/nmrium/blob/5c0556dd98ba73301b76729825c47d20350c3f2e/src/data/SpectraManager.ts#L104-L137

The serializer should be able to receive directly the NMRium state. If it is not the case it is the serializer that should be improved.

lpatiny (changes requested):

src/component/NMRium.tsx
hamed-musallam: @lpatiny 

the problem the View and data are in the same state but Settings is not

lpatiny (approved):
Look's ok to me

Issue #2166, opened 2023-02-20
Fix SVG export to be compatible with word


jschneidewind:
# Current behaviour

Currently it is possible to export vector graphics of spectra as `.svg`. While this file format is well suited for many applications, there are some issues when using it with Microsoft Word, where it often does not render properly.

See for example (tested with Microsoft Word for Mac 2019):
[MLB-170-1_10_Word_SVG_import.pdf](https://github.com/cheminfo/nmrium/files/10783714/MLB-170-1_10_Word_SVG_import.pdf) (spectrum is only partially rendered and integral values are missing)

In contrast to how it should look like in the `.png`: 
![MLB-170-1_10](https://user-images.githubusercontent.com/41544471/220093022-8bb6afef-7a42-409a-887c-2022eec05cc2.png)

A work-around is to convert the `.svg` to `.pdf` using software such as Inkscape. In this conversion, however, some graphic elements are lost (e.g. the integral values). 
See for example: [MLB-170-1_10_Inkscape_SVG_to_PDF.pdf](https://github.com/cheminfo/nmrium/files/10783731/MLB-170-1_10_Inkscape_SVG_to_PDF.pdf)

Furthermore, when the color of the spectrum is changed within NMRium and it is then saved as `.svg`, the spectrum disappears in the `.svg` to `.pdf` conversion (see for example: [MLB-170-1_10_Color_Changed_Inkscape_SVG_to_PDF.pdf](https://github.com/cheminfo/nmrium/files/10783796/MLB-170-1_10_Color_Changed_Inkscape_SVG_to_PDF.pdf)). The spectrum is also not visible when the `.svg` is directly imported into Word.

# Suggested feature

If possible, it would be great if an option for the direct export to `.pdf` is added, which would expand the options for vector graphic export and enable direct use with Microsoft Word (which is of course commonly used for both teaching and publications).

Thanks for your work on NMRium!

For reference here are the original `.zip` Bruker file as well as the `.svg` files with original and changed color:

[MLB-170-1.zip](https://github.com/cheminfo/nmrium/files/10783857/MLB-170-1.zip)
![MLB-170-1_10](https://user-images.githubusercontent.com/41544471/220088946-bfb94765-00f9-48e6-a53b-19eee6abeeb4.svg)
![MLB-170-1_10_Color_Changed](https://user-images.githubusercontent.com/41544471/220092284-d9d11b60-ca5d-4276-8a27-b481bec107e3.svg)

[...]

PR #2204, opened 2023-03-01
fix: use svg transform instead of css transform

hamed-musallam:
https://github.com/cheminfo/nmrium/issues/2166

### DIFF
--- a/src/component/1d/Chart1D.tsx
+++ b/src/component/1d/Chart1D.tsx
@@ -1,4 +1,5 @@
 import SpectrumInfoBlock from '../1d-2d/components/SpectrumInfoBlock';
+import { usePreferences } from '../context/PreferencesContext';
 
 import ApodizationLine from './ApodizationLine';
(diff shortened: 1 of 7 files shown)


lpatiny:
So with auto you use something different for display and SVG export ?

Please rename Shape rendering to 'Spectra rendering'

image

hamed-musallam:
> Please rename Shape rendering to 'Spectra rendering'
> 
> image

it is not only for spectra, i set it over the root element 
https://github.com/cheminfo/nmrium/blob/f4681581849b8a82227b70c1e58e32a57007e8df/src/component/1d/Chart1D.tsx#L30
https://github.com/cheminfo/nmrium/blob/f4681581849b8a82227b70c1e58e32a57007e8df/src/component/2d/Chart2D.tsx#L43

> So with auto you use something different for display and SVG export ?

No, by default the `auto` is the default value for shape rendering.

PR #2212, opened 2023-03-02
chore: reducing some label size

lpatiny:


hamed-musallam:
We did the changes in https://github.com/cheminfo/nmrium/pull/2216

Issue #2211, opened 2023-03-02
800 x 600px goal


lpatiny:
We would like to be able to process spectra even on 800x600 px screen.

I changed by hand the size of the component to test

![image](https://user-images.githubusercontent.com/1484241/222473541-cb1f612c-fd6d-45e3-a393-9f04511aee12.png)

The main issues are in the top bar and bottom bar

![image](https://user-images.githubusercontent.com/1484241/222474158-3b9bd6cd-1a39-45d1-b984-012ce22fa462.png)

For the bottom should be easy to fix, if there is not enough room we remove intensity and index, for 1D only delta is usefull.

![image](https://user-images.githubusercontent.com/1484241/222474429-94d674bb-d767-4acf-8c5b-0e0c1d9c59d3.png)

For 2D only delta (F1 and F2) are required

![image](https://user-images.githubusercontent.com/1484241/222474640-d046a272-5277-4f34-a8af-acf7eab5bbda.png)

For apodization if there is not enough room we can use abbreviation

![image](https://user-images.githubusercontent.com/1484241/222474794-ed3760fb-eac7-4197-916f-c3d649ee9d70.png)

- LB for 'Line broadening'
- GB for 'Gauss broadening'
- LB center for 'lineBroadeningCenter'

For phase correction the fields with the values could be small and we should only display the first 2 decimals. Mouse over could give the full label.

Some other tools need to be adapted as well so that they fit in 800px wide.

PR #2216, opened 2023-03-03
style: adapt header and footer to work on small container size

hamed-musallam:
closes: https://github.com/cheminfo/nmrium/issues/2211


lpatiny:
We could remove the Select workspace and save workspace if the screen is too small

image

For phase correction we can reduce the size of the manual number input box

image

But we need to keep at least a button for PH0 and PH1 so that we can click on it and drag to change the phase with the mouse. The label on small screen could just be PH0 and PH1.

For baseline correction the degree is a number between 1 and 9 so the input field could always be smaller

image

Issue #2164, opened 2023-02-17
Testcases for source


lpatiny:
In the `data` state we have the `spectra` property but also the `source` property.

The `source` property is defined IF the data source can be saved. This means that we can save the data source rather than the data itself. This is an option in the 'Save as...' feature of NMRium.

![image](https://user-images.githubusercontent.com/1484241/219783039-d140beac-1762-495d-b92f-2c74ccccc1e9.png)

We need to test the following possibilities:

##  We load a file like  '1H spectrum test' (click in the menu)

![image](https://user-images.githubusercontent.com/1484241/219783204-43cfc7b6-b56b-4093-898e-18d8ce747836.png)

If we 'save as...'

![image](https://user-images.githubusercontent.com/1484241/219783372-357d840c-2c8e-4d76-b544-32622e9bf206.png)

Raw data, Data Source and No Data should be active !!! (the 3, not like in this capture)

- If we save `Raw Data`, the source property is not present and there is data in a spectrum
- If we save `Data Source`, the source property is present and there is no `data` in a spectrum but there is in 'selector' with the property 'files'
- If we save `No data` the data property is not present.

### After this we drag / drop a new file

If we 'save as...'

![image](https://user-images.githubusercontent.com/1484241/219783372-357d840c-2c8e-4d76-b544-32622e9bf206.png)

Raw data No Data should be active !!! (Only 2, Data source is not available)

- If we save `Raw Data`, the source property is not present and there is data in a spectrum
- If we save `No data` the data property is not present.

## Load data from a drag / drop

If we 'save as...'

![image](https://user-images.githubusercontent.com/1484241/219783372-357d840c-2c8e-4d76-b544-32622e9bf206.png)

Raw data No Data should be active !!! (Only 2, Data source is not available)

- If we save `Raw Data`, the source property is not present and there is data in a spectrum
- If we save `No data` the data property is not present.

wadjih-bencheikh18:
@lpatiny when we save `Raw Data` we have the same `source` property as `Data Source`

```js
{
  entries: [
    {
      baseURL: '',
      relativePath: 'data/cytisine/1H_Cytisin_600MHz-R+I.dx'
    }
  ]
}
```

PR #2230, opened 2023-03-11
Testcases for source

wadjih-bencheikh18:
closes: https://github.com/cheminfo/nmrium/issues/2164

### DIFF
--- a/test-e2e/panels/spectra.test.ts
+++ b/test-e2e/panels/spectra.test.ts
@@ -483,3 +483,74 @@ test('Export source from 1H spectrum', async ({ page }) => {
     expect(data).toBeUndefined();
   });
 });
+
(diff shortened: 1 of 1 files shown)


wadjih-bencheikh18:
@lpatiny I mentioned in the issue
https://github.com/cheminfo/nmrium/issues/2164#issuecomment-1464921279

The problem why the test is failing

wadjih-bencheikh18:
> `autoassignment` test is failing too i'm trying to figure out the reason

I fixed this

wadjih-bencheikh18:
We still have this error

when we save `Raw Data` we have the same `source` property as when we save `Data Source`

```js
{
  entries: [
    {
      baseURL: '',
      relativePath: 'data/cytisine/1H_Cytisin_600MHz-R+I.dx'
    }
  ]
}
```

PR #2231, opened 2023-03-13
feat: integrate with fifo-logger

hamed-musallam:


lpatiny:
Please remove the 'zero' if there is no bugs

image

Also if there is no 'bug' and there are some warning message it could be displayed as orange + number

And if there are some 'info' messages (and no bugs and no warning) it could be in green with the number of messages.

hamed-musallam:
@lpatiny 

We could have a mix of logs with different Levels, I mean the number of references to all kinds of logs, so based on what you need to re-change the color of the badge

lpatiny:
the badge will be the color of the more critical event (highest error level) and only this one would be displayed

hamed-musallam:
@lpatiny 

Then this means we have to order logs highest level to the lower ones and then count the logs for the highest level.

in this case, we should have color for each level to change the badge color based on it or we could use the ones that you purposed in another ticket 

>Light background color for the line could highlight the levels:

>pink for error and fatal
>lightyellow for warn
>lightgreen for info, debug and trace

lpatiny:
We don't need to order all the logs. We should need to filter / count each level and then sort per level to check what is the more serious problem.

PR #2233, opened 2023-03-13
fix: parse hex color and extract opacity

hamed-musallam:
- [X] https://github.com/cheminfo/nmrium/issues/2166#issuecomment-1440513307

We parse the color to extract the color in hex and the alpha and convert the alpha hex value to a decimal value between 0 and  1.

https://github.com/cheminfo/nmrium/blob/2cdf4dd99c16f44f1434a495d8941ef1619a6be0/src/component/utility/parseHexaColor.ts#L1-L20

This should solve the problem of why the spectra are not visible when the import the exported SVG  to MS World

### DIFF
--- a/src/component/1d/Line.tsx
+++ b/src/component/1d/Line.tsx
@@ -4,6 +4,7 @@ import { useScaleChecked } from '../context/ScaleContext';
 import useActiveSpectrumStyleOptions from '../hooks/useActiveSpectrumStyleOptions';
 import useXYReduce, { XYReducerDomainAxis } from '../hooks/useXYReduce';
 import { PathBuilder } from '../utility/PathBuilder';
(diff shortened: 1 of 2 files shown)


hamed-musallam:
@lpatiny 

WDT ? is it good for you to parse the color in with way ?

lpatiny:
Exporting to SVG and importing to word is ok:

image

However if from word I print to PDF I obtain:

image

How can I solve this issue ?

hamed-musallam:
I think this issue is related to how the MS Word PDF conversion works, currently, The MS Word version that i use for the PDF exportation works perfectly!!!

https://github.com/cheminfo/nmrium/issues/2166#issuecomment-1467632043

lpatiny:
Yes we should merge this PR. It opens correctly in inkscape.

PR #2240, opened 2023-03-16
refactor: Improve logging system

hamed-musallam:
- [x] In the general preference add an option on the first tab with 'logging level' (from trace to error). Default value 'info'.
- [x] For the 'badge' display only the number of unread messages
- [x] In the list remove the column 'Level' (keep only label)
- [x] In the list remove the column 'meta'
- [ ] In the list for the error replace by a 'copy' button. It is expected to be a stack trace if there is an error
- [x] For the 'date' just keep hoour:minute:second. You can use
- [x] Add a bullet in front of 'new' logs (the one that were not yet available last time you open the log window)
- [ ] When opening the dialog, scroll to the first 'new' log.
- [x] https://github.com/cheminfo/nmrium/issues/2242

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "clipboard-polyfill": "^4.0.0",
         "d3": "^7.8.2",
         "eventemitter3": "^5.0.0",
-        "fifo-logger": "^0.5.0",
+        "fifo-logger": "^0.6.0",
         "file-saver": "^2.0.5",
         "filelist-utils": "^1.7.1",
         "formik": "^2.2.9",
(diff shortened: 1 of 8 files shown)


lpatiny:
> We use the sequential id to recognize the last one seen and because you define a new logger object for each new context we lose the sequential id because it is a private member of the class and not static

This should be fixed now.

lpatiny:
> we use the objectInspector component I mean you can inspect the object directly there, is this enough or do you still need to have a copy button

We can still decide this later. The idea is that if there is an error maybe it should be an information to report to us but it is not clear yet.

lpatiny:
> I think it is better to order the loges based on 'time' descending because all the users will look for the recent logs usually

I was thinking about other applications that would have this behaviour and indeed a mail application sorts the mails more recent first so you may solve this issue using the sorting as you suggested.

Issue #1898, opened 2022-11-11
Export current spectrum as jcamp if 1D


lpatiny:
If there is one spectrum selected and it is a 1D there could be one more export possibility:

Export current spectrum as JCAMP-DX

There is more or less a good example here but we should take care if it is a FID or a FT file.

https://github.com/cheminfo/convert-to-jcamp/blob/master/src/__tests__/from1DNMRVariables.test.ts#L39-L84

@hamed-musallam Could you try to implement this ?

This will require as dependency `convert-to-jcamp`

It will use: [from1DNMRVariables](https://github.com/cheminfo/convert-to-jcamp/blob/master/src/from1DNMRVariables.ts)

targos:
Do you think it's more intuitive than right click -> export as jcamp directly in the list?

lpatiny:
I was thinking about it. Maybe we could allow both.

lpatiny:
Or it is easier to implement we can only have the export as JCAMP-DX with right click on the spectrum

PR #2283, opened 2023-04-03
Export current state as JCAMP-DX

hamed-musallam:
- [x] https://github.com/cheminfo/nmrium/issues/1898


src/data/SpectraManager.ts
```diff
@@ -123,3 +130,56 @@ export function toJSON(
...
+    jcamp = toJcamp1D(spectrum);
+  } else {
+    throw new Error('convert 2D spectrum to JCAMP is not supported');
+  }
+
+  const blob = new Blob([jcamp], { type: 'text/plain' });
+  saveAs(blob, `${spectrum.info.title}.jdx`);
+}
+
+function toJcamp1D(spectrum: Datum1D) {
```
lpatiny, line 146: Please move this code to nmr-load-save. It will also be easier to add a testcase.

The dependency convert-to-jcamp should also be only in nmr-load-save.

You may expose the new method `spectrum1DToJcamp`

lpatiny:
Exportation should only be available on 1D spectra.

Please rename to 'Export current state as JCAMP-DX'

Alejandro will remove all meta information but frequency, solvent, pulse sequence and nucleus.

lpatiny:
@jobo322 It does not seem standard to export in the JCAMP-DX

```
##nucleus=1H
##pulseSequence=zg30
##baseFrequency=600.2
##originFrequency=600.202658313
##isFid=false
```

##label is 'normalized' and defined and we should not invent new fields. What is the reason they are there ?

jobo322:
> @jobo322 It does not seem standard to export in the JCAMP-DX
> 
> ```
> ##nucleus=1H
> ##pulseSequence=zg30
> ##baseFrequency=600.2
> ##originFrequency=600.202658313
> ##isFid=false
> ```
> 
> ##label is 'normalized' and defined and we should not invent new fields. What is the reason they are there ?

yes it is a mistake, those values are not used to load the jcamp at the end. 

in convert-to-jcamp
info object is passed to LDR like `##label` and the meta object as `##$label`, I propose only write upperCase variables that come from `info` object so, with it we will still have a degree of freedom from nmr-load-save to add new LDRs without modify `convert-to-jcamp` package

hamed-musallam:
@jobo322 

these lines of code should not be here and i think they should be moved to nmr-load-save 
https://github.com/cheminfo/nmrium/blob/25ed85ed36a775ca4c26b7acd1fdd4d20add6606/src/data/SpectraManager.ts#L132-L185

Issue #2235, opened 2023-03-14
Experimental FT for 2D spectra


lpatiny:
@hamed-musallam 

Could you prepare new buttons for 2D processing:

- Zero filling
- Apodization
- FT

Those 3 buttons will need options (the 3)

For Zero filling we will need 2 zero filling, for first and second dimensions
For apodization we will also need the options for the 2 dimensions
For FT we should be able to select if we want to make FT in both dimensions or not.

All those buttons should be hidden behind the experimental features and this will help @jobo322 to implement the code.

hamed-musallam:
Do you think it is better to wait until we see what options that Alejandro needs for each filter before we add those tools(buttons) in the main toolbar

lpatiny:
Can indeed just be a toolbar with 'apply' 'cancel' 'preview' but currently no options.

![image](https://user-images.githubusercontent.com/1484241/224989951-2fa117ec-9d99-4a70-8586-4585bf12454c.png)

hamed-musallam:
@lpatiny after I checked, maybe we do not need to create a different tool for 2d, we can use the sample 1D tools and based on Nucleus we apply the 1D filter or the 2D filter, WDT?

hamed-musallam:
closed by https://github.com/cheminfo/nmrium/pull/2294

PR #2294, opened 2023-04-12
feat: 2d Fourier transform

hamed-musallam:
- #2235


public/data/brukerFolders/cosyAndHsqcFid.json
targos, line 1: This file is wrong. It contains typed arrays converted to JSON.

jobo322: thanks, it should not contain data, many thanks.

targos: How did you generate it?

jobo322: It was generated by reading the FileColection from a folder, I forgot to remove the data and info from the nmrium returned file. So It was the reason it appear like that. I forgot too to upload the correct one

targos: Ok, but how exactly? I wonder, because this nmrium file you uploaded here is wrong. When the data is included, it should be as normal arrays (like `[12, 13, 14]`, not `{"0": 12, "1", 13, "2", 14}`), so there may be a bug somewhere when the JSON is generated.

jobo322: yes it is true, it were due I did a JSON.stringify to convert the json into string. Due I was expecting to have not data I did not add the function to convert the typed array into normal array.

PR #2296, opened 2023-04-13
fix: onChange callback debounce

hamed-musallam:
refactor input component

### DIFF
--- a/src/component/elements/Input.tsx
+++ b/src/component/elements/Input.tsx
@@ -51,6 +51,17 @@ const styles: Record<
   },
 };
 
+function check(value, type) {
+  if (type === 'number') {
+    const pattern = /^(?:-?\d*|\d+)(?:\.\d{0,20})?$/;
+    if (value.trim() === '' || pattern.test(value)) {
+      return true;
+    }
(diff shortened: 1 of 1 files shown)


src/component/elements/Input.tsx
```diff
@@ -107,20 +122,20 @@ const Input = forwardRef(
...
+    } = props;
     const [val, setVal] = useState(value);
     const valueRef = useRef(value);
     const localRef = useRef<any>();
     const combinedRef = useCombinedRefs([ref, localRef]);
 
-    const debounceOnChange = useRef(
-      debounce((value) => {
-        onChange(value);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
targos, line 132: This is the most important rule in a React project. If you have to disable it, it means you are leaving a bug in your code.

hamed-musallam: I think we could useMemo instead of useCallback which returns a function 
- #2297

targos: Thanks, that PR looks good to me.

Issue #2313, opened 2023-04-20
Assignment and floating molecule


lpatiny:
We should be able to easily assign a range to many atoms and this seems not currently possible.

![2023-04-20 20 51 37](https://user-images.githubusercontent.com/1484241/233460561-5f2af8a2-6d3d-4763-8d90-539751a951b4.gif)

hamed-musallam:
-  #2322

Issue #2311, opened 2023-04-20
Assignment issues


lpatiny:
Should the minus icon not remove assignment ?

Currently it add the number in the sum column. Little weird.

![2023-04-20 20 44 42](https://user-images.githubusercontent.com/1484241/233459228-451433c7-e51e-4b9a-9637-fa3617f49fcf.gif)

PR #2322, opened 2023-04-26
fix: assignments 

hamed-musallam:
- [X] #2313
- [X] #2311


hamed-musallam:
@lpatiny 

in the old code, there is something strange for me why do we have to remove the range nbAtoms if we remove the signals assignments, currently, i stopped this code because the range could be assigned to atoms different than the one that the signals have.

https://github.com/cheminfo/nmrium/blob/c4a91f1096d5dcce9663e41f91816ee07b15193f/src/data/utilities/RangeUtilities.ts#L80

Issue #2309, opened 2023-04-20
Can open preferences only once


lpatiny:
When you open the preferences and then press 'ESC' you can not open the preferences any more.

![2023-04-20 20 21 53](https://user-images.githubusercontent.com/1484241/233454241-6b98ca68-d72f-439e-9930-2ae171eaeebb.gif)

targos:
Fixed in the latest version of react-science.

targos:
Fixed by https://github.com/cheminfo/nmrium/commit/fde906ee3233b736b24e9c7eb6b52af13b1e1c19

PR #2328, opened 2023-04-28
2309 can open preferences only once

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/2309


targos:
Thanks, but the issue is already fixed on the main branch. The changes in package-lock don't look right to me here.

PR #2329, opened 2023-04-28
Refactor AlignSpectraModal using react-science modal

wadjih-bencheikh18:


lpatiny:
Because we need some interaction with the spectrum in order to select from / to, Modal is not the right approach.

We propose to replace the content of the panel rather than creating a modal.

targos:
@wadjih-bencheikh18 Please fix the conflicts.

wadjih-bencheikh18:
> @wadjih-bencheikh18 Please fix the conflicts.

Refactored

PR #2339, opened 2023-05-05
check symmetry first for manual range picking

jobo322:
the symmetry check will be manage by multiplet-analysis package and we will use signalJoinCouplings to merge couplings, e.g. the pattern `dd` with closer coupling values into `t` based in a `tolerance` option.

### DIFF
--- a/src/component/modal/MultipletAnalysisModal.tsx
+++ b/src/component/modal/MultipletAnalysisModal.tsx
@@ -155,8 +155,13 @@ export default function MultipletAnalysisModal({
       try {
         const result = analyseMultiplet(analysesProps, {
           frequency: info.originFrequency,
+          minimalResolution: 0.3,
+          maxTestedJ: 17,
           takeBestPartMultiplet: true,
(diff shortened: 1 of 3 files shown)


lpatiny:
This looks much better than before. The idea is to check first if the multiplet is symmetric before applying the multiplet analysis right ?

There is a related bug: https://github.com/cheminfo/nmrium/issues/2340 but this can be solved after.

lpatiny:
Could you also check the range picking on a 13C. Because it fails. It never finds out it is a singulet.

targos:
Please take the time to write at least title to your pull requests.

jobo322:
> Please take the time to write at least title to your pull requests.

I agree, I will continue adding a description for each PR.

jobo322:
> This looks much better than before. The idea is to check first if the multiplet is symmetric before applying the multiplet analysis right ?
> 
> There is a related bug: #2340 but this can be solved after.

actually it is done at the moment to merge coupling constants. It is based in a tolerance option.

PR #2352, opened 2023-05-11
use a central source of possible patterns for multiplicities

jobo322:


lpatiny:
Indeed multiplicity should come in the future from nmr-processing

Issue #2359, opened 2023-05-18
Spectra and data source


lpatiny:
When saving .nmrium file we can have the `data.source` property

```
{
  "version": 4,
  "data": {
    "source": {
      "baseURL": "",
      "entries": [
        {
          "relativePath": "data/multiSpectra.zip"
        }
      ]
    },
```

However in the current version I don't see any reference to the source at the level of each spectrum (a selector).

I'm convinced that this was the case before and it disappear and of course this add bugs.

For instance if you load this set of spectra:

![image](https://github.com/cheminfo/nmrium/assets/1484241/c4362cf9-d357-4df2-94d4-71a45292c315)

And delete some of them then save the 'data source' in a .nmrium file you can not reload correctly the .nmrium file.

Is this a format migration problem ?

The missing property is `sourceSelector`

PR #2360, opened 2023-05-18
fix: spectrum object should include the sourceSelector

hamed-musallam:
closes: https://github.com/cheminfo/nmrium/issues/2359


lpatiny:
It works much better indeed. If I take the following dataset

https://us06web.zoom.us/j/5370637587?pwd=cnBYNWI0dHptSGlVajBDc2xVRlBOdz09

We see that for the 13C we have a FID (not processed spectrum) which is ok.

However if I save as .nmrium and reload again the spectrum it has been processed and this should not happen.

PR #2370, opened 2023-05-19
Show time domain spectrum if none of the selected spectrum is FT

jobo322:


lpatiny (approved):
Functionality is ok for me.

Issue #1771, opened 2022-09-25
General preferences dialog box should be centered vertically


lpatiny:
This issue is related to https://github.com/zakodium-oss/analysis-ui-components/issues/218 and the new dialog component should be reused (if possible) in this project.

When you click in the top right to open general preferences

![image](https://user-images.githubusercontent.com/1484241/192130499-da5067e5-fff0-4f26-83f2-c073b5a93c71.png)

- top banner so that it is similar to the one of 'about'

![image](https://user-images.githubusercontent.com/1484241/192130518-7b0cfc48-0af1-466e-8957-aeeaf02e381f.png)

- like for the 'about' dialog box it should be centered vertically

- the current size is too small on large sreen (HD).

I would try the following rules:
- minimal size is 600 x 400
- maximal size is 1000 x 800
- the width should not be larger than 80% of NMRium component width
- the height should not be larger than 80% of NMRium component height

@stropitek does those rules make sense ? I would not like to have super complex code just to deal with the dialog size.

stropitek:
@wadjih-bencheikh18 it should be possible to use the Modal component from analysis-ui-components with the latest release

The component was designed such as either
- The modal size adapts to the content + maxWidth -> should be use for static content
- The modal width / height is fixed -> should be used for dynamic content

PR #2377, opened 2023-05-24
Fix modals style and general preferences

wadjih-bencheikh18:
closes: https://github.com/cheminfo/nmrium/issues/1771


targos (changes requested):
The goal of `react-science` is to take care of the style of common components as much as possible. We should do general changes to the style of modals in `react-science` so it benefits other projects and doesn't make nmrium more complex.
Also I think in this case, the styles for the body of the about dialog (color of links, bullet points, size of image, etc) is very specific and should not be applied to other modals.
Can you please open a PR on `react-science` with a subset of the changes that we can review?

wadjih-bencheikh18:
> Can you please open a PR on `react-science` with a subset of the changes that we can review?

Here's the PR that will allow us to change header text position in `react-science` modal 
https://github.com/zakodium-oss/react-science/pull/512

targos:
What is the modal for which you want to change the header position?

wadjih-bencheikh18:
> What is the modal for which you want to change the header position?

I have to change the position of general preferences modal

targos:
I have trouble to understand what this PR does. I see it adds "maxWidth" property, but there is no difference between this PR and https://dev.nmrium.org (main branch), whatever is the size of the browser window.

wadjih-bencheikh18:
> I have trouble to understand what this PR does. I see it adds "maxWidth" property, but there is no difference between this PR and https://dev.nmrium.org (main branch), whatever is the size of the browser window.

I followed what Daniel mentioned here https://github.com/cheminfo/nmrium/issues/1771#issuecomment-1285741723 to check if there any difference. Now i'm trying to change modal content style so it will adapt screen size

PR #2380, opened 2023-05-26
feat: rename metaInfo to customInfo and reorder information panel

lpatiny:
@hamed-musallam 

I rename the field 'metaInfo' to 'customInfo' to avoid some confusions.

I also added in the cytisin example some 'customInfo':

```json
        "customInfo": {
          "values": [{"key": "solvent", "value": "DMSO-d6"}, {"key": "temperature", "value": "298.15 K"}]
        },
```

However:
- the customInfo is not loaded from the .nmrium file
- we need to check that when saving the .nmrium file those customInfo are really saved


hamed-musallam:
>        "customInfo": {
          "values": [{"key": "solvent", "value": "DMSO-d6"}, {"key": "temperature", "value": "298.15 K"}]
        },
        
        the structure that you use was wrong, it should be like bellow 
        
        `
        "customInfo": {
          "solvent": "DMSO-d6",
          "temperature": "298.15 K"
        },
        `

PR #2408, opened 2023-06-09
fix: spectra prediction

hamed-musallam:


lpatiny:
If you take a 1D cytisine and try to predict everything it crash for the 2D

image

hamed-musallam:
> If you take a 1D cytisine and try to predict everything it crash for the 2D
> 
> image

this happening because we did not pass the molfile correctly and I fix it right now, maybe this happening as a result of the last refactoring I can not be sure

Issue #2358, opened 2023-05-18
Simplify prediction


lpatiny:
1. Remove the Add button

![image](https://github.com/cheminfo/nmrium/assets/1484241/f49ef80e-1ad2-4bd6-90ae-b233cbb38555)

2. We can only have one predicted spectrum

Remove all the spectra that has pulse sequence 'prediction' before adding new spectra

3. In the prediction panel have directly the editor

4. If in the prediction 'workspace' have the floating molecule active

hamed-musallam:
@lpatiny 

i think it is better if you change your proposed structure to `predictions: Record<moleculeId,string[]>`
because it is supposed to have one set of predicted spectra

hamed-musallam:
>Ideally we should also display the Line Width (is it stored in info ???).

we are not storing the `line width` in the info object, we just have it in the view object

lpatiny:
@jobo322 Would it make sense to add `lineWidth` in info ? Especially in the case of predicted spectra it is an important property of the spectrum. Actually for the prediction what is the shape of the peaks ? We could also think about the possibility to customize the shape.

jobo322:
I think the shape of the peaks should be in ranges, also there should be the FWHM of each peak. I could storage the lineWidth in the info object for each predicted spectra

PR #2419, opened 2023-06-12
refactor: simplify the prediction

hamed-musallam:
- [X] https://github.com/cheminfo/nmrium/issues/2358

### DIFF
--- a/src/component/panels/MoleculesPanel/MoleculeHeader.tsx
+++ b/src/component/panels/MoleculesPanel/MoleculeHeader.tsx
@@ -22,7 +22,6 @@ const styles: Record<'toolbar' | 'labelInput', CSSProperties> = {
     borderBottom: '0.55px solid rgb(240, 240, 240)',
     padding: '5px 10px',
     justifyContent: 'space-between',
-    height: '35px',
   },
   labelInput: {
     width: '150px',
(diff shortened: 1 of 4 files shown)


lpatiny:
Replace prediction is great. Could we also add a second button 'Add prediction'^

image

Replace would replace the last added prediction.

Can we have the edition mode even when it is empty ? If I click on the 'prediction' workspace I still need to click to call the editor

image


Could we have a more compact design ? Font could be smaller and also 'Line width' could be next to 'Frequency'

image

Most of the time we will not change those preferences.

Please remove this 'Auto save' message. Not sure why it is there.

image

Is there a simple way to have a local state for the editor and that the displayed molecule as floating is related to the prediction ? Currently we can edit the molecule but the displayed spectrum is possibly not the one corresponding to the molecule

hamed-musallam:
> Add prediction 

is it going to append the new predicted spectra to the existing ones?

>Please remove this 'Auto save' message. Not sure why it is there.

I added the message to notify the user that the change is synced locally, i will remove it.

>Could we have a more compact design ? Font could be smaller and also 'Line width' could be next to 'Frequency'

yes, we can do that and make the form compact.

>Is there a simple way to have a local state for the editor and that the displayed molecule as floating is related to the prediction ? Currently we can edit the molecule but the displayed spectrum is possibly not the one corresponding to the molecule

i did not get what you need me to do. can you clarify things for me in another way

hamed-musallam:
@lpatiny 

`StructureEditor` onChange callback not triggering from the first click once we re-render the editor, you need to trigger twice

lpatiny:
Functionality looks great ! You may merge it when it is ready.

PR #2427, opened 2023-06-12
add missing multiplicity in the solvent databases

jobo322:
solvent databases lacks of some multiplicities like m or Br t that exists in the .tsv source file.


jobo322:
this fix will be included in:

- https://github.com/cheminfo/nmrium/pull/2386

Issue #2429, opened 2023-06-12
unhandled error in signal detection


targos:
```
length of signal should not exceed 2048 points

{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/reducer/Reducer.js',
  line: 324,
  column: 47,
  name: 'draft'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/reducer/actions/RangesActions.js',
  line: 272,
  column: 0,
  name: 'addNewRange'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/reducer/actions/RangesActions.js',
  line: 255,
  column: 16,
  name: 'data'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/data/data1d/Spectrum1D/ranges/addRange.js',
  line: 24,
  column: 19,
  name: 'detectSignal'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/data/data1d/Spectrum1D/ranges/detectSignal.js',
  line: 55,
  column: 0,
  name: null
}
```

hamed-musallam:
fixed by #2438

PR #2438, opened 2023-06-15
fix: auto (peaks and ranges) picking  window size

hamed-musallam:
- [X] #2429
- [X] https://github.com/cheminfo/nmrium/issues/2434


lpatiny:
I don't think adding a range manually is useful. 

You may delete completely the code of the '+' because once added we are in trouble to try to edit this range. Better to brush the range.

image

Maybe it was added in https://github.com/cheminfo/nmrium/pull/1566 but no idea why.

Also when you click on the 'edit' button you could systematically change the active tool to the 'range picking' so that you can resize immediately the range.

Issue #2432, opened 2023-06-12
unhandled error in CrossLinePointer


targos:
```
Cannot read properties of undefined (reading 'step')

{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/1d-2d/tools/CrossLinePointer.js',
  line: 36,
  column: 0,
  name: 'step'
}
```

hamed-musallam:
fixed by https://github.com/cheminfo/nmrium/pull/2439

Issue #2431, opened 2023-06-12
unhandled error in footer banner


targos:
```
Cannot destructure property 'startX' of '(0 , I.useContext)(...)' as it is undefined.

{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/1d/FooterBanner.js',
  line: 80,
  column: 12,
  name: null
}
```

hamed-musallam:
closed by https://github.com/cheminfo/nmrium/pull/2439

PR #2439, opened 2023-06-15
refactor: brush and mouse context

hamed-musallam:
the bugs happen because the `DOWN` action in the brush reducer does not return the state on certain conditions.

 - [X] #2431
 - [X] #2432

### DIFF
--- a/src/component/1d-2d/tools/BrushX.tsx
+++ b/src/component/1d-2d/tools/BrushX.tsx
@@ -1,6 +1,6 @@
-import { CSSProperties, useContext } from 'react';
+import { CSSProperties } from 'react';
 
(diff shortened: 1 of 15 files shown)


lpatiny:
We could not reproduce the issue but it is likely in the following line:

https://github.com/cheminfo/nmrium/blob/8fa41502c09c26436d339e1be891acb0da2dcffd/src/component/EventsTrackers/BrushTracker.tsx#LL261C7-L261C7

You don't return 'state' so it could be undefined

Please also add the return type to this method

https://github.com/cheminfo/nmrium/blob/8fa41502c09c26436d339e1be891acb0da2dcffd/src/component/EventsTrackers/BrushTracker.tsx#L205

That would have allow us to see this issue.

PR #2464, opened 2023-06-26
fix: integral value editing if the sum 0

hamed-musallam:


targos:
Does it fix this error:

```
Cannot read properties of undefined (reading 'options')

{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/reducer/actions/IntegralsActions.js',
  line: 16,
  column: 8,
  name: 'setSumOptions'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/data/data1d/Spectrum1D/SumManager.js',
  line: 33,
  column: 0,
  name: 'options'
}
```

lpatiny:
If you go through zero and then set a value from a line the sum is NaN

PR #2475, opened 2023-06-26
fix simulation of only negatives chemical shift

jobo322:
- handle errors in generate spectrum from ranges
- handle errors in join couplings

prevent crash when a prediction (carbon or proton) is missing.


targos:
Can you update your PR title to something meaningful? I doubt this is adding a crash.

targos:
Included in https://github.com/cheminfo/nmrium/pull/2488

PR #2484, opened 2023-06-30
fix: validate JCAMP file URL correctly

hamed-musallam:


targos:
What about this point from the issue:

> IMO we shouldn't do that. Any URL can return a JCAMP, it doesn't have to end with the right extension.

hamed-musallam:
the problem in this case we can not know that this is jcamp so we can not handle it in the correct way unless we can detect the binary signature but I'm not sure if the jcamp files have a specific signature .

targos:
This should be handled by nmr-load-save, no?

hamed-musallam:
yes, but even in nmr-load-save we look for the extension https://github.com/cheminfo/nmr-load-save/blob/3134061c7d6ad4be434bb0f3da3f8f76891b3a3a/src/read/read.ts#L160-L162

hamed-musallam:
WDT? can we merge this PR?

targos:
I see. Yes we can merge, and maybe think about a way to force nmr-load-save to interpret files as jcamp in the future.

Issue #2406, opened 2023-06-09
Replace with checkbox component from react-science


lpatiny:
Currently you can not click on the label of the checkbox but you need to click on the checkbox itself.

This can be solve using the react-science checkbox component

![2023-06-09 09 57 32](https://github.com/cheminfo/nmrium/assets/1484241/3a2c3a81-1531-4cae-8f81-f5015cd56d8a)

hamed-musallam:
i think no need to use the checkbox from react-science, we can wrap the `IsotopesViewer` with a label or we can modify the IsotopesViewer to render it as a label.

https://github.com/cheminfo/nmrium/blob/main/src/component/elements/IsotopesViewer.tsx

targos:
We don't "need" to use components from react-science, but the goal with that project is to build a good component library that is separate from nmrium to be more easily testable, discoverable, and to have more external contributions in order to improve it.

wadjih-bencheikh18:
> Currently you can not click on the label of the checkbox but you need to click on the checkbox itself.

It's currently working in [https://dev.nmrium.org](https://dev.nmrium.org) , i checked the `FormikCheckBox` component and it works perfectly
@targos we can maybe update `Checkbox` component in `react-science` and make it more customizable (styling, positions ...)

targos:
It was probably fixed after opening this issue, but the goal here is to replace uses of the [CheckBox](https://github.com/cheminfo/nmrium/blob/main/src/component/elements/CheckBox.tsx) component with the one from `react-science` (so in the end we can delete the component).
If you think we need to add features to the `react-science` component, please comment here with examples.

wadjih-bencheikh18:
the feature that need to be add to use it in nmrium is checkbox style (label and input)
the style is used here in [FormikCheckBox](https://github.com/cheminfo/nmrium/blob/main/src/component/elements/formik/FormikCheckBox.tsx)

targos:
Can you start by migrating all the other uses of the component (skip FormikCheckBox)?
Before adding props to the API, we should understand why they are needed. The decision can be to remove the option because it's not necessary.

wadjih-bencheikh18:
We still have to
use react-science `CheckBox` in `FormikCheckBox`

targos:
Yes, but no. I would like to push so Formik itself is replaced with something else (probably react-hook-form as that's what we have experience with). The Formik project is dead.

hamed-musallam:
> The Formik project is dead

[...]

PR #2495, opened 2023-07-04
2406 replace with checkbox component from react science

wadjih-bencheikh18:
- [x] use `react-science` CheckBox in `PredictSpectraModal`, `FiltersTable` and `ZeroFillingOptionsPanel`
- [x] fix Live changes functionality in `ZeroFillingOptionsPanel`
- [ ] use `react-science` CheckBox in `FormikCheckBox`

Refs: https://github.com/cheminfo/nmrium/issues/2406

### DIFF
--- a/src/component/header/ZeroFillingOptionsPanel.tsx
+++ b/src/component/header/ZeroFillingOptionsPanel.tsx
@@ -1,11 +1,11 @@
 import { NmrData1D } from 'cheminfo-types';
 import { Filters } from 'nmr-processing';
 import { useEffect, useState, memo } from 'react';
+import { Checkbox, CheckedState, Field } from 'react-science/ui';
 
 import generateNumbersPowerOfX from '../../data/utilities/generateNumbersPowerOfX';
 import { useDispatch } from '../context/DispatchContext';
 import ActionButtons from '../elements/ActionButtons';
-import CheckBox from '../elements/CheckBox';
 import Label from '../elements/Label';
 import Select from '../elements/Select';
 import { useFilter } from '../hooks/useFilter';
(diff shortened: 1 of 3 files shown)


src/component/header/ZeroFillingOptionsPanel.tsx
```diff
@@ -83,17 +77,9 @@ function ZeroFillingOptionsInnerPanel(props: { size: number }) {
...
-      >
-        <CheckBox
-          name="livePreview"
-          defaultChecked
-          onChange={checkChangeHandler}
-        />
-      </Label>
+      <Field label="Live preview" name="livePreview">
+        <Checkbox checked={livePreview} onChange={setLivePreview} />
+      </Field>
```
targos, line 82: Should not use `Field`, but `label` prop on the chekbox (it will put the text on the right, but I think it's better). Also I think you can remove the "name"

targos (changes requested):

src/component/modal/PredictSpectraModal.tsx
```diff
@@ -149,8 +145,8 @@ function PredictSpectraModal({
           spectra for confidential molecules.
         </p>
         <div className="warning-container">
-          <CheckBox
-            onChange={approveCheckHandler}
+          <Checkbox
+            onChange={setApproved}
             checked={isApproved}
             key={String(isApproved)}
           />
```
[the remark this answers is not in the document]
wadjih-bencheikh18: Okay, i will fixe it. It just that the label looked better before the checkbox

src/component/header/ZeroFillingOptionsPanel.tsx
wadjih-bencheikh18: Okay, i will fixe it. It just that the label looked better before the checkbox

Issue #1537, opened 2022-05-06
Use Dialog from blueprintjs instead of `modal.show`


lpatiny:
@hamed-musallam Could you check if the `modal.show` can be replaced using the Analysis-ui-components ?

https://github.com/redhaam/analysis-ui-components/blob/ac8b11cd12dfb0b32e09e409760ca0bcb1be1007/src/layout/Modal.tsx#L15-L16

If there are any limitations we should improve analysis-ui-components.

hamed-musallam:
@lpatiny 

the current modal have limitation  and I think it should be discussed with Michael, the modal  in anaylsis-ui-compoennt is injected under RootLayout component and this will not work with NMRium because we need the modal deeply under all provider, and this is why we inject it at this level https://github.com/cheminfo/nmrium/blob/3865f572881a09d796d93de6a121b37c5d6d0ca3/src/component/NMRium.tsx#L362-L375

targos:
I need a concrete example of something that doesn't work if the modal is injected elsewhere.

hamed-musallam:
@targos 

as I can see the modal is injected directly under the root layout component, does in the current version the modal can be injected elsewhere?

targos:
No it cannot but I don't see what is the issue with it

hamed-musallam:
in many cases, in the NMrium we access the provider's hooks and if we inject and with the current implementation of the modal this is not possible because it is on a different level in the tree. is this right?

targos:
The only thing that matters is where the modal is rendered/open inside nmrium, not where the portal will be in the dom

targos:
I'm going to migrate one modal in nmrium to show the new API

lpatiny:
@hamed-musallam Michael is finishing the example and after you may refactor the code based on the example. 
For now there will be less options and the modal will always be centred on the screen and the background will always be disabled.

lpatiny:
@wadjih-bencheikh18 Please make one PR per dialog !

wadjih-bencheikh18:
@targos 
Is it possible to use context hook to migrate modals?
it just creating a context hook similar to the one we already have but using `react-science` Modal
So we will be able to call useModal hook to use any modal easily
```
const modal=useModal()
modal.show(<ModalComponent {...props} />
```
the modal can be closes inside `ModalComponent` and we may do it using context.

Because right now I don't really like the way I'm migrating, we must have a button and it toggle the modal. And it's not always the case

[...]

PR #2505, opened 2023-07-07
refactor PredictSpectraModal using react-science

wadjih-bencheikh18:
Refs: https://github.com/cheminfo/nmrium/issues/1537

### DIFF
--- a/src/component/modal/PredictSpectraModal.tsx
+++ b/src/component/modal/PredictSpectraModal.tsx
@@ -40,6 +40,7 @@ const styles = css`
       margin: 5px 5px 5px 0;
     }
   }
+
   .header {
     padding: 0;
     margin: 0 30%;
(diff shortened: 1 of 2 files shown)


targos (changes requested):
The modal should be closed when you click the predict button

wadjih-bencheikh18:
We still have one problem which is that the loading dialog must be above the modal, I tried to change `zIndex` but it doesn't fix it

targos:
To me, the problem is that we should not use a separate loading alert if the modal stays open. Instead, we should either:
- Keep the loading alert, but close the modal immediately 
- Remove the loading alert and display something inside the modal

targos:
There's something wrong with your Prettier

wadjih-bencheikh18:
> There's something wrong with your Prettier

fixed. I had a problem in my vscode settings

targos (approved):

PR #2507, opened 2023-07-10
fix: autoExtendRange option and logs for simulation with signals are out of range.

jobo322:


lpatiny:
Ok for me the PR works correctly. What you could still do maybe is more precise to tell that proton was not predicted because there are none or because there is only labile proton.

package-lock.json
targos, line 1: Please cancel all changes to this file.

jobo322: is it correct now @targos ?

targos: yes

PR #2508, opened 2023-07-11
refactor database structure search modal

wadjih-bencheikh18:
refs : https://github.com/cheminfo/nmrium/issues/1537


targos (changes requested):
Please compare with the current version on https://dev.nmrium.org

- Modal title isn't nice (there's a grey background only around the text)
- Molecule editor is cropped

CleanShot 2023-07-11 at 15 38 29

targos:
Also, this shouldn't be implemented as a modal, because the results are displayed live in the Database panel while you edit the molecule.

I will discuss this with @lpatiny to suggest a replacement.

targos:
We will keep this on hold for now. @wadjih-bencheikh18 feel free to migrate other uses of Modal.

lpatiny:
@targos So the current idea is that we have a large Modal that also display the selected molecule 'live'. Is this correct ?

But what will happen when we draw the molecule. Will the panel be filtered live as well ? (and we can see it being filtered in the greyed panel ?

image

lpatiny:
For now we will loose the interaction and we should merge this PR.

Please rebase the project to have a final check ?

wadjih-bencheikh18:
> Please compare with the current version on https://dev.nmrium.org

Design fixed
image

lpatiny (approved):
Functionality is ok for me.

stropitek (approved):

targos:
There's a conflict.

Issue #2481, opened 2023-06-28
other unhandled error in auto ranges detection


targos:
```
GSD only accepts monotone increasing x values

{
  source: 'webpack://_N_E/node_modules/nmrium/lib/component/reducer/actions/RangesActions.js',
  line: 76,
  column: 12,
  name: 'detectRanges'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/data/data1d/Spectrum1D/ranges/detectRanges.js',
  line: 7,
  column: 38,
  name: 'spectrum'
}
{
  source: 'webpack://_N_E/node_modules/nmrium/lib/data/data1d/Spectrum1D/ranges/autoRangesDetection.js',
  line: 37,
  column: 38,
  name: null
}
{
  source: 'webpack://_N_E/node_modules/nmr-processing/lib-esm/xy/xyAutoRangesPicking.js',
  line: 8,
  column: 16,
  name: 'xyAutoPeaksPicking'
}
{
  source: 'webpack://_N_E/node_modules/nmr-processing/lib-esm/xy/xyAutoPeaksPicking.js',
  line: 33,
  column: 0,
  name: 'getPeakList'
}
{
  source: 'webpack://_N_E/node_modules/nmr-processing/lib-esm/xy/xyAutoPeaksPicking.js',
  line: 58,
  column: 21,
  name: 'data'
}
{
  source: 'webpack://_N_E/node_modules/ml-gsd/lib-esm/gsd.js',
  line: 19,
  column: 0,
  name: null
}
```

PR #2511, opened 2023-07-12
fix: unexpected errors in reducer crashing the app

tpoisseau:
try-catch the reducer and set errorAction in state on catch Also I typed the Alert API from useAlert / AlertContext

Closes: https://github.com/cheminfo/nmrium/issues/2481


tpoisseau:
https://github.com/cheminfo/nmrium/pull/2516
Done here

PR #2516, opened 2023-07-17
chore: log error from spectrum reducer

tpoisseau:


hamed-musallam:
if we are going to report the error with the logger then i think no need to show it with the alert dialog, the second thing, i think we have to take care of is to erase the `errorAction` or/and we should be sure that the `useEffect` will be triggered only on the case the `errorAction` changed.

did you test this mechanism in action? to be sure that things work well

tpoisseau:
I did not flush error after log, to avoid unnecessary re-render.
I didn't found how to produce error in specter reducer to test it but I'm confident it's working as expected

src/component/NMRium.tsx
```diff
@@ -455,12 +455,14 @@ export default memo(NMRium);
 function StateError() {
   const { errorAction } = useChartData();
   const { error: alertError } = useAlert();
+  const { logger } = useLogger();
 
   useEffect(() => {
     if (!errorAction) return;
 
     alertError(errorAction?.message ?? String(errorAction));
```
hamed-musallam, line 463: No need to use the alert because the logger modal will popup automatically if the log level is warning, error, or fatal

tpoisseau: use logger instead alert done

hamed-musallam:
Thanks, no more comments from my side

PR #2528, opened 2023-07-20
fix: limiting the vertical zoom

hamed-musallam:


src/component/reducer/helper/Zoom1DManager.ts
```diff
@@ -14,6 +14,10 @@ export const ZOOM_TYPES = {
 
 export type ZoomType = keyof typeof ZOOM_TYPES;
 
+function toFixed(value: number) {
+  return Number(value.toFixed(7));
```
targos, line 18: There's no guarantee that it will stay with this precision when converted back to a number.
What problem does this PR attempt to fix?

hamed-musallam: I found if the value fraction is too small (1e-7 ... etc) this makes a problem in the plotting and the spectrum is no longer visible in the other hand it does not make sense to zoom to infinity.

targos: Converting to a string is certainly not the way to fix this issue.

hamed-musallam: we should first put a limitation( hard code value) on the max value for the domain to prevent the infinity zoom but it still does not know what should be the value and based on what

lpatiny: The problem that Hamed tries to fix is that if you scale too much the spectrum disappear.



The reason for me is not clear. what are the values of the array when it disappear ? Is there a NaN in the array or an infinity ?

I would expect a 'Math.min' or 'Math.max' somewhere to fix the issue if it is related to an infinity or NaN.

hamed-musallam: I still need to do more testing to see exactly when this bug happen and why and this why and this why i set the status of the PR as pending

hamed-musallam: we limit the vertical zoom-in based on the spectra max value and multiply it with 0.0001

targos (changes requested):
new behavior doesn't make sense to me. the zoom stops arbitrarily

src/component/reducer/helper/Zoom1DManager.ts
hamed-musallam: Why we do that because if we zoom in for infinity some of the scaled values it too big or too small that the path element cannot display it, for example `-2.12222222e+10` so i set limitation on the zoom in

hamed-musallam:
this happened because we set a limit how much you zoom

lpatiny:
I checked the SVG (played with values in the path) and apparently the values in a path are float 32. In the specification it is 'float': https://www.w3.org/TR/SVG11/paths.html

A variable of type float32 can store decimal numbers ranging from 1.2E-38 to 3.4E+38. This is the range of magnitudes that a float32 variable can store. This applies to negative and positive numbers.

When we get this kind of 'error':

```Error: <path> attribute d: Expected number, "…7.7549115913556 -4.1258572051422…".```

It is related to a number being larger (or smaller) than a float32.

From this information it should be possible based on the max and min value to calculate the limit of the possible scaling and to block the scaling to this limit (well we could go to 90% of the max / min).

By applying this we can just not make better.

targos:
I need time to understand the problem at hand. The values in the SVG do not have to be the same as the values in the spectrum. D3 provides tools to rescale values before passing them to SVG.

hamed-musallam:
the D3 does its job to rescale the spectrum but in order to zoom in/out we increase/decrease the domain and as a result of that the domain becomes too big or too small and so do the scaled values.

more details about the limits of numbers in SVG

https://oreillymedia.github.io/Using_SVG/extras/ch08-precision.html

targos:
I propose https://github.com/cheminfo/nmrium/pull/2551 as an alternative (tested on Chrome, Safari and Firefox. zoom is not limited and the spectrum never disappears)

PR #2554, opened 2023-07-28
feat: show label under floating molecules

hamed-musallam:
Extracting the label and exporting it with the SVG is not an ideal solution. Instead, we should consider forking the 'react-rnd' library which uses two other packages `react-draggable` and `react-resizer`, and modifying it to work seamlessly with SVG elements.


lpatiny:
There is a PR that should simplify the addition of the label:
- https://github.com/zakodium-oss/react-ocl/pull/24

lpatiny:
It would be nice that displaying the label is an option as well. Not sure where to place it. It could be another toggle button next to floating.

lpatiny:
The label position and copy is now ok. 

I don't know how to toggle show / hide the label.

I also think a single click should be enough to edit the label. Also it is not clear how to valide / cancel the change. Do we have a component to do this that would display small X or V ?

lpatiny:
I think the molecule panel could have his own settings with:
- show / hide label
- labelFontSize
- labelColor

https://github.com/cheminfo/nmrium/blob/649d0afda5ad7b7bfdb7b9eb335cadc1d20b2e25/src/component/1d-2d/components/FloatMoleculeStructures/DraggableStructure.tsx#L134-L136

PR #2559, opened 2023-07-31
refactor: shortcut for zero filling and FFT

hamed-musallam:


lpatiny:
Indeed it works as expected, thanks !

PR #2560, opened 2023-07-31
Migrate MultipletAnalysisModal

wadjih-bencheikh18:
refs : https://github.com/cheminfo/nmrium/issues/1537


lpatiny:
Not sure what is the current status of this PR.

When the dialog is open we still need to be able to interact in the spectrum in order to select new chemical shifts or new coupling constants while pression shift pressed and drawing on the spectrum.

wadjih-bencheikh18:
I worked on `Analyse Multiplet` Modal

image

src/component/1d/Viewer1D.tsx
```diff
@@ -387,6 +381,16 @@ function Viewer1D({ emptyText = undefined }: Viewer1DProps) {
                     </MouseTracker>
                   </BrushTracker>
                 )}
+              {brushDataRef.current && (
```
targos, line 384: I haven't looked into it, but something is wrong here. You should never use a ref's value during rendering, because changing the ref doesn't trigger a rerender.

wadjih-bencheikh18: fixed, I used a state.

lpatiny:
@wadjih-bencheikh18 Ok thanks for the migration. Please could you increase the default size ? I would like we avoid if possible the horizontal scroll.
Vertically it could also be higher.

image

targos:
If you remove the fixed size, it should automatically use as much as necessary. If not, that's a bug in react-science

wadjih-bencheikh18:
When I resize the window, I don't know why the plots changes. Maybe because the scale is changing.
[Screencast from 16-08-23 12:40:36.webm](https://github.com/cheminfo/nmrium/assets/56769876/bc7d22e3-16f7-4842-a173-aac913d7304b)

targos:
There's a conflict. Apart from that, is the PR ready?

wadjih-bencheikh18:
> Apart from that, is the PR ready?

Yes the PR is ready
image

targos:
E2E tests are broken.

PR #2562, opened 2023-08-01
feat: add new way to display peaks

hamed-musallam:


lpatiny:
You may invert the logic so that by default it displays the peak values on the top and when you click on the toggle it would display the value just over the peak.

hamed-musallam:
Is the edit mode still necessary in the new display mode? It appears that you intend to set it as the primary display mode

src/component/panels/PeaksPanel/PeaksPanel.tsx
```diff
@@ -207,6 +208,21 @@ function PeaksPanelInner({
           >
             <SvgNmrPeaks style={{ pointerEvents: 'none', fontSize: '12px' }} />
           </ActiveButton>
+
+          <ActiveButton
+            style={{ marginLeft: '2px', marginRight: '2px' }}
+            popupTitle={
+              peaksViewState.displayingMode === 'group'
+                ? 'Single Mode'
+                : 'Group mode'
```
targos, line 217: From a user's perspective, I'm not sure these two names make much sense. In the new mode, the peaks don't look like they are grouped. It's more like "spread".

hamed-musallam: actually i did not find a good name this is why i use `group`, any suggestion it will be helpful

lpatiny:
The tests are not passing but the functionality looks ok to me.

src/component/elements/NumberInput.tsx
```diff
@@ -29,13 +37,22 @@ const NumberInput = forwardRef(
       step = 'any',
       min = 'any',
       max = 'any',
+      enableAutoSelect = false,
```
targos, line 40: `autoSelect` seems enough and more simple

targos: I wonder if we should have instead `selectOnFocus`, which would be more intuitive than selecting when the component is rendered. It's useless anyway to select the contents if the element is not focused.

targos: well, it seems that `.select()` also takes the focus, which is not obvious by reading the MDN doc/examples. So we can keep `autoSelect` for simplicity.

PR #2594, opened 2023-08-08
fix: 2d shift filter

hamed-musallam:


hamed-musallam:
> You're fixing a critical bug. Don't use "chore" for the commit message.

I will  squash and merge this PR so the name of the PR i prefixed it with fix

targos:
well, be careful when you merge, because by default GitHub takes the message of the commit.

PR #2599, opened 2023-08-10
feat: add integrals from Bruker 1D data

jobo322:


lpatiny:
Current the integrals appear between square brackets:

image

The reason is that there is no property `kind` for each integral.

@jobo322 Please at the level of nmr-load-save if there are some 'integrals' always add the property `"kind": "signal"`


```json
            {
              "from": 8.379,
              "to": 8.001,
              "integral": 2.57899,
              "id": "9026937e-75b3-48a1-9d7b-fa05602b3011",
              "originFrom": 8.379,
              "originTo": 8.001,
              "absolute": 2412006.0881226463,
              "kind": "signal"
            },
```

lpatiny:
It is nice you added 1i and 1r for aspirin. 

Please also add a new example: `File collection with integrals` just bellow

image

Issue #2590, opened 2023-08-08
Test 2D reference change


lpatiny:
In one of the existing 2D tests, we likely have 'zones' (using the Zone tool 'picking').

In the zone panel, we are able to change the value of a zone along X (first column) and Y (second column).

Currently, this crashes, but we need to check that after editing this value, there is actually this new value in the list.
We should be able to edit the first value as well as the second value (second column). This corresponds to the value in the X and Y axis.

![2023-08-08 06 53 11](https://github.com/cheminfo/nmrium/assets/1484241/8f78beb7-bcd4-401a-b86f-9aef9d1e047d)

PR #2612, opened 2023-08-11
2D spectra reference change

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/2590

### DIFF
--- a/src/component/2d/zones/Signal.tsx
+++ b/src/component/2d/zones/Signal.tsx
@@ -59,10 +59,7 @@ function Signal({ signal, isVisible }: SignalProps) {
     <g className="zone-signal">
       {isVisible.signals && (
         <g>
-          <SignalCrosshair
-            signal={signal}
-            key={`${signal.x.delta},${signal.y.delta}`}
-          />
+          <SignalCrosshair signal={signal} />
           <circle
             onMouseEnter={() => {
               assignment.show();
--- a/src/component/2d/zones/Signal.tsx
+++ b/src/component/2d/zones/Signal.tsx
@@ -59,7 +59,10 @@ function Signal({ signal, isVisible }: SignalProps) {
     <g className="zone-signal">
       {isVisible.signals && (
         <g>
-          <SignalCrosshair signal={signal} />
+          <SignalCrosshair
+            signal={signal}
+            key={`${signal.x.delta},${signal.y.delta}`}
+          />
           <circle
             onMouseEnter={() => {
               assignment.show();
--- a/test-e2e/panels/ranges.test.ts
+++ b/test-e2e/panels/ranges.test.ts
@@ -398,7 +398,7 @@ test('2D spectra reference change', async ({ page }) => {
       nmrium.page.locator(
         `_react=Signal[signal.x.delta=${x}][signal.y.delta=${y}]`,
       ),
-    ).toHaveCount(1);
+    ).toBeVisible();
   });
   await test.step('Change reference', async () => {
     await nmrium.page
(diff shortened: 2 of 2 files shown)


test-e2e/panels/ranges.test.ts
```diff
@@ -354,3 +357,67 @@ test('Auto peak picking on all spectra', async ({ page }) => {
...
+    await expect(
+      nmrium.page.locator(
+        '_react=ZonesPanel >> _react=ZonesTableRow >> nth=0 >> td >> nth=1',
+      ),
+    ).toHaveText(x.toFixed(2));
+    await expect(
+      nmrium.page.locator(`_react=SignalCrosshair[key=/^${x}\\d*,${y}\\d*/]`),
+    ).toHaveCount(1);
+    await expect(nmrium.page.locator('_react=XAxis')).toHaveText(xAxisDefault);
+    await expect(nmrium.page.locator('_react=YAxis')).toHaveText(yAxisDefault);
```
wadjih-bencheikh18, line 421: The tests fail because of this

src/component/2d/zones/Signal.tsx
```diff
@@ -59,7 +59,10 @@ function Signal({ signal, isVisible }: SignalProps) {
     <g className="zone-signal">
       {isVisible.signals && (
         <g>
-          <SignalCrosshair signal={signal} />
+          <SignalCrosshair
+            signal={signal}
+            key={`${signal.x.delta},${signal.y.delta}`}
```
targos, line 64: React key should not be used for tests, because it has real effects on React rendering

wadjih-bencheikh18: Would you please suggest a way to get delta values in test

Also I have a question , can regex be used with CSS locators as well?
In the documentation is used only on react component, and it didn't work in the other cases

targos: The values you are interested in are in the `signal` prop. Can't you match using the nested property syntax?
The example from the docs should help you: https://playwright.dev/docs/other-locators#react-locator

targos: I seems you can use regex in CSS selectors. They mention it in https://playwright.dev/docs/other-locators#css-matching-by-text

wadjih-bencheikh18: > Can't you match using the nested property syntax?

I couldn't do that because in the signal prop we have the exact value of the delta but in the panel we have the rounded value

targos: but in the key you don't put the rounded value here

wadjih-bencheikh18: > I seems you can use regex in CSS selectors.

Regex here is used the textual content, I used it to match props [https://playwright.dev/docs/other-locators#react-locator](https://playwright.dev/docs/other-locators#react-locator)

wadjih-bencheikh18: > I couldn't do that because in the signal prop we have the exact value of the delta but in the panel we have the rounded value

I used the exact value to test, and it's working now thank you

targos: I think you need to discuss with @lpatiny, because I don't understand what is important to be tested in this case.

targos: What happens when you try a regex with the numerical value?

wadjih-bencheikh18: > What happens when you try a regex with the numerical value?

I didn't get any result

I even tried 
```tsx
await expect(
      nmrium.page.locator(
        '_react=Signal[signal.x.delta=/\\.*/][signal.y.delta=/\\.*/]',
      ),
    ).toHaveCount(1);
```
and I got 
```tsx
Expected: 1
Received: 0
```

test-e2e/panels/ranges.test.ts
wadjih-bencheikh18: @lpatiny the tests are failing because of a bug

When we change the delta of any zone the axis scale changes too

PR #2617, opened 2023-08-15
fix: correct the integration sum for integrals and ranges

### LINKED ISSUE #2601: Integral behaviour is buggy
If you toggle between `Fix integral values` we get some strange behaviour.

You can correctly integrate twice, then toggle and integrate one more time. At this level you should have switch to the manual mode to define the sum of the integrals but it does not seem the case.

If you again toggle the button you start to get crazy values.

Please also check for the ranges if it is ok.

hamed-musallam:
- [x] create a new release of nmr-processing

close #2601

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "multiplet-analysis": "^2.1.2",
     "nmr-correlation": "^2.3.3",
     "nmr-load-save": "^0.17.1",
-    "nmr-processing": "^10.0.4",
+    "nmr-processing": "^10.0.5",
     "nmredata": "^0.9.2",
     "numeral": "^2.0.6",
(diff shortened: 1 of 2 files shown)


lpatiny (approved):

package.json
```diff
@@ -88,7 +88,7 @@
     "multiplet-analysis": "^2.1.2",
     "nmr-correlation": "^2.3.3",
     "nmr-load-save": "^0.17.1",
-    "nmr-processing": "^10.0.1",
+    "nmr-processing": "10.0.3-pre.1692096567",
```
hamed-musallam, line 91: waiting for a new release of nmr-processing

PR #2618, opened 2023-08-15
Use react-science confirmation dialog

wadjih-bencheikh18:
I couldn't migrate modals in `KeysListenerTracker` and `FilterTable` because the modal have 3 options, so we will able to use `react-science` after improving [`ConfirmModal`]( https://github.com/zakodium-oss/react-science/issues/525)


src/component/panels/IntegralsPanel/IntegralPanel.tsx
```diff
@@ -195,6 +193,34 @@ function IntegralPanelInner({
...
+            display: 'flex',
+            flex: '1 1 0%',
+            alignItems: 'center',
+            justifyContent: 'center',
+            fontWeight: 'bold',
+            margin: 25,
+            fontSize: 14,
+            color: 'rgb(175, 0, 0)',
+          }}
+        >
```
targos, line 220: This is not ok. You're repeating the same styles over and over.
Either we accept the new style, or we improve the style of the component in react-science.

wadjih-bencheikh18: Okay I will try to remove it

wadjih-bencheikh18: With styling: 
image

Without:
image

I think we should update `ConfirmModal` style in r`eact-science`

wadjih-bencheikh18: the needed style
```tsx
style={
            fontWeight: 'bold',
            margin: 25,
            fontSize: 14,
            color: 'rgb(175, 0, 0)',
}
```

targos: No, make a PR to `react-science` (with stories if needed)

lpatiny:
If we press Enter when the dialog is there we get a focus around the dialog and this should not be the case.

image

targos:
This PR will need to be updated now that the styling issues are fixed in react-science.

wadjih-bencheikh18:
> This PR will need to be updated now that the styling issues are fixed in react-science.

The outline problem is solved but there is an issue in the new release
https://github.com/zakodium-oss/react-science/issues/575

wadjih-bencheikh18:
This PR need to be closed

PR #2622, opened 2023-08-16
fix: spectra calibration 

hamed-musallam:


lpatiny:
I would put the message in red directly in the dialog box. What do you think ? But the functionality is ok.

PR #2652, opened 2023-08-30
feat: improve signal kinds for ranges and integrals

hamed-musallam:


lpatiny:
When renaming ranges become undefined.

PR #2654, opened 2023-08-31
feat: display peaks in ranges

hamed-musallam:
Tasks regarding the current implementation 

- [x] Take care of the shift value for the peaks inside the ranges.
- [x] Delete peak from the range 
- [x] hide/show peaks  within ranges


hamed-musallam:
@lpatiny 


I think the current way of handling peaks, where some are standalone peaks and others are within defined ranges without any interconnection might not be the best way to do it.

I propose implementing a way to link between the peaks found within the defined ranges and those identified as standalone. This link can be established by associating each peak with a rangeKey and signalKey, indicating its connection with a specific range > signal, or relying on a unique peak key.

if we agree on this, we should update in both ways between standalone peaks and those within the ranges whenever a range/signal is added, deleted, edited, or when a standalone peak is added, which may or may not be associated with a range.

One option is to keep the current structure of organizing things with ranges > signals > peaks. Alternatively, we could simplify it by putting all peaks, whether they're on their own or linked, into one list. This would prevent us from having duplicate peak information in both the standalone peaks and those that are inside the ranges.

PR #2658, opened 2023-09-05
util: add utility to calculateDerivedProperties

lpatiny:
This should simplify the process to add properties when spectrum is shifted based on the original data.

You should chedk the test to determine how it works


hamed-musallam:
Released as an independent package, https://github.com/cheminfo/derived-props

PR #2659, opened 2023-09-05
fix: jcamp exportation ntuples

jobo322:
pre-release nmr-load-save


jobo322:
nmr-load-save will be updated in #2660.

PR #2670, opened 2023-09-14
chore: update to ts522

lpatiny:
The issue with ml-spectra-processing is solved but we still have other issues:

```
src/component/toolbar/ToolBar.tsx:2:25 - error TS2307: Cannot find module 'react-science/ui' or its corresponding type declarations.

2 import { Toolbar } from 'react-science/ui';
                          ~~~~~~~~~~~~~~~~~~

src/utils/clipboard/clipboard.ts:2:10 - error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'.

2   return import('clipboard-polyfill').then((c) => c.readText());
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/utils/clipboard/clipboard.ts:6:10 - error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'.

6   return import('clipboard-polyfill').then((c) => c.read());
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/utils/clipboard/clipboard.ts:10:10 - error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'.

10   return import('clipboard-polyfill').then((c) => c.writeText(data));
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/utils/clipboard/clipboard.ts:21:10 - error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'.

21   return import('clipboard-polyfill').then((c) => c.write(data));
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            ```

@targos Could you advice ?

Does it make sense otherwise to remove the publication of commonJS ?


targos (approved):
nmrium.org is ok with this change

hamed-musallam (approved):
I can confirm that it also works well on my end

Issue #2676, opened 2023-09-17
General preferences not used to generate ACS string


lpatiny:
<img width="674" alt="image" src="https://github.com/cheminfo/nmrium/assets/1484241/d8ae04e8-3482-4a61-a64c-0132e408f100">

<img width="574" alt="image" src="https://github.com/cheminfo/nmrium/assets/1484241/aa986e20-4c60-4e06-971a-8338188ccc4e">

You can use numeral in nmr-processing. Would make probably sense.

https://github.com/cheminfo/nmr-processing/blob/2c64c30fef1754f044222dbd0308f958ddb06f32/src/ranges/rangesToACS.ts#L17-L22

jobo322:
the formatting preferences differ from ranges preferences.

jobo322:
also we should refactor the name of formatting preferences 
<img width="440" alt="image" src="https://github.com/cheminfo/nmrium/assets/12143935/cc042915-7914-4a26-8331-4711ff8e9ad6">

ppm make sense but Hz for couplings is not clear

PR #2684, opened 2023-09-18
correct ranges-preferences to generate acs string

jobo322:


src/component/panels/RangesPanel/RangesHeader.tsx
```diff
@@ -137,16 +138,11 @@ function RangesHeader({
...
-      );
-      const nbDecimalJ = getNumberOfDecimals(rangesPreferences.deltaHz.format);
-
+      const currentPreferences = preferences.nuclei[nucleus.toLowerCase()];
       const result = rangesToACS(ranges.values, {
         nucleus, // '19f'
-        nbDecimalDelta, // 2
-        nbDecimalJ, // 1
+        nbDecimalDelta: currentPreferences.ppm,
+        nbDecimalJ: currentPreferences.hz,
```
hamed-musallam, line 145: I have some notes:

1- we format them based on the delta in ppm and Hz from the panel preferences which is per nucleus.

Screenshot 2023-09-19 at 09 26 14

why do you need to take the format from the general preferences?

hamed-musallam: As i know the`nbDecimalDelta` and `nbDecimalJ` are accepting a number of decimals and not the format

hamed-musallam (changes requested):

src/component/panels/RangesPanel/RangesHeader.tsx
jobo322: it was a confusion based on the description of the ticket #2676, also the use of deltaHz for coupling decimals is wrong because we have Coupling(Hz)

hamed-musallam: does it accept both `number` or `string`?

src/component/panels/RangesPanel/RangesHeader.tsx
```diff
@@ -138,15 +138,10 @@ function RangesHeader({
...
-      const nbDecimalDelta = getNumberOfDecimals(
-        rangesPreferences.deltaPPM.format,
-      );
-      const nbDecimalJ = getNumberOfDecimals(rangesPreferences.deltaHz.format);
-
       const result = rangesToACS(ranges.values, {
         nucleus, // '19f'
-        nbDecimalDelta, // 2
-        nbDecimalJ, // 1
+        nbDecimalDelta: rangesPreferences.deltaPPM.format,
```
lpatiny, line 143: This property should be renamed. Forget about the breaking change in nmr-processing.
- deltaFormat
- jFormat

hamed-musallam (approved):

src/component/panels/RangesPanel/RangesHeader.tsx
jobo322: deltaFormat and couplingFormat was the names

PR #2692, opened 2023-09-21
fix: support nmredata extension

jobo322:
- .nmredata files as a sdf instead of a nmrRecord.
- .nmredata and .sdf file extension will be checked if contains nmredata information.


public/data/nmredata/nmredata.json
```diff
@@ -4,18 +4,22 @@
       "baseURL": "",
       "entries": [
         {
-          "name": "menthol_1D_1H_assigned_J.nmredata",
-          "size": 1,
-          "relativePath": "data/nmredata/menthol_1D_1H_assigned_J.nmredata",
-          "lastModified": 0
+          "relativePath": "data/nmredata/menthol_1D_1H_assigned_J.zip"
```
lpatiny, line 7: Please revert. It should change nothing.

PR #2671, opened 2023-09-15
feat: 2D phase correction

hamed-musallam:
- [x] select pivot based on the sum (same code as add missing projections)
- [x] IF phase2D is selected, scroll wheel scale the 1D spectra (the traces)
- [x] IF phase2D is selected, SHIFT + scroll wheel scale the 2D (optional)
- [x] If over a trace, backspace will delete the trace (highlight of a zone of 10 pixels)
- [x] Selecting the pivot using SHIFT + click
- [x] The traces should share the same vertical domain
- [x] For 1D phase correction change the behaviour so that pivot is SHIFT + click
- [x] For vertical phase correction the spectra has to be vertical.
- [x] apply phase correction over the traces.

PR #2696, opened 2023-09-26
refactor: scale ratio function

hamed-musallam:


hamed-musallam:
rebased with 2d-phase-correction branch #2671

PR #2699, opened 2023-09-28
feat: vertical zoom by panning vertically with two fingers

hamed-musallam:


targos:
I tried on iPhone and it doesn't work at all. It zooms way too fast vertically and at the same time it also zooms horizontally. This is unusable.

hamed-musallam:
I initially assumed that onTouchStart would be fired before onPointerDown when interacting with a touch-enabled device. However, it appears that this is not the case, and the flag I defined to prevent horizontal scrolling is ineffective.

This issue highlights the necessity for thorough testing on actual devices. Unfortunately, I am currently facing difficulties connecting my phone to the development environment, preventing me from testing and rectifying the touch event handling in our application.

i need to resolve the connection issue then i will return back to this PR

hamed-musallam:
I will close it for now

PR #2712, opened 2023-10-17
fix: build esm, not cjs

stropitek:


src/utils/clipboard/clipboardHooks.ts
```diff
@@ -199,9 +199,7 @@ export function useClipboard(): UseClipboardReturn {
       async rawWriteWithType(data: string, type = 'text/plain') {
         try {
           const item = await newClipboardItem({
-            [type]: new Promise((resolve) => {
-              resolve(new Blob([data], { type }));
-            }),
+            [type]: new Blob([data], { type }),
```
stropitek, line 202: I don't know what the impact of this change is.

I had a TS error pop after changing the TS configuration

stropitek: CleanShot 2023-10-17 at 14 32 55

targos:
Seems to work fine for nmrium.org

hamed-musallam (approved):

hamed-musallam:
Seems it works as well for the other projects

src/utils/clipboard/clipboardHooks.ts
targos: https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/ClipboardItem#data

stropitek: ok I can look into the type error.

@hamed-musallam can you check if the feature linked to this code still works in safari?

https://github.com/cheminfo/nmrium/commit/a6b54c956a7b3c8b860ecd346174879d91df769c

stropitek: But not related to this change right?

It also does not work on https://dev.nmrium.org/

hamed-musallam: it seems it is a bug, it should work as far I remember, on the other hand, the copy, and paste to the clipboard works well with the text.

hamed-musallam: i think it is not related to your changes

stropitek: CleanShot 2023-10-18 at 15 42 36

This is a comment added by the polyfill next to the relevant type. `ClipboardItemData` is the type that includes `Promise<string | Blob>`

https://github.com/w3c/clipboard-apis/pull/126

PR #2730, opened 2023-10-30
add processing filters from nmr-load-save

jobo322:


jobo322:
the correct changes are included in 
- https://github.com/cheminfo/nmrium/pull/2737

PR #2772, opened 2023-11-29
fix: avoid error in nmredata parsing

jobo322:
improve speed by parallel I/O requests


hamed-musallam:
done with https://github.com/cheminfo/nmrium/pull/2770

PR #2777, opened 2023-12-01
refactor: display small buttons for each panels when not enough room

hamed-musallam:


lpatiny:
Seems this is related to https://github.com/zakodium-oss/react-science/issues/243 in which we wanted to create icons for each panel.

Issue #2781, opened 2023-12-01
Printing spectrum


lpatiny:
I would like that if we try to print the page containing a spectrum we only display what is usually in the clipboard (when we copy)

![image](https://github.com/cheminfo/nmrium/assets/1484241/08b6084b-fd73-4ea6-bf43-8957b90a7ac7)

We can use to achieve this goal the `@media print` css rule but currently the way the dom is organised makes it very difficult.

Ideally we would need a div id="nmrium-viewer-content' that contains all that should be printed.

This will likely require that the bottom toolbar is moved.

![image](https://github.com/cheminfo/nmrium/assets/1484241/5f925079-9569-4434-b588-2a7edda3598f)

PR #2786, opened 2023-12-04
feat: print spectra

hamed-musallam:
- #2781

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "ml-tree-similarity": "^2.2.0",
         "multiplet-analysis": "^2.1.2",
         "nmr-correlation": "^2.3.3",
-        "nmr-load-save": "^0.34.2",
+        "nmr-load-save": "^0.35.0",
         "nmr-processing": "^12.8.0",
         "nmredata": "^0.9.11",
         "numeral": "^2.0.6",
@@ -9551,9 +9551,9 @@
       }
     },
     "node_modules/nmr-load-save": {
-      "version": "0.34.2",
-      "resolved": "https://registry.npmjs.org/nmr-load-save/-/nmr-load-save-0.34.2.tgz",
-      "integrity": "sha512-8L1NXJ+75HK1AB4ooSu9pKiFBwJmom2OMcoJAsNRFX5W6tX7AUidt47k13xUtAu3LRvqKGXIwAi/9PWW4ZBgpA==",
+      "version": "0.35.0",
+      "resolved": "https://registry.npmjs.org/nmr-load-save/-/nmr-load-save-0.35.0.tgz",
+      "integrity": "sha512-uur725neK10hdrbyVJYArG4wSUSBT1raXQCgByWvGyirDLyJ330vsOn/ilRv1kKIV4i4twX84aJPZA2Eyh0F6w==",
       "dependencies": {
         "@lukeed/uuid": "^2.0.1",
         "@types/lodash.merge": "^4.6.9",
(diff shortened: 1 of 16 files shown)


lpatiny:
@hamed-musallam 

Could you finalize this PR ?

The goal is that when you print from the browser fill either an A4, B4, Letter (?) page depending of your general preferences.

It could be related to: https://stackoverflow.com/questions/16649943/css-to-set-a4-paper-size

Currently we have in the bottom the grey banner that should not be present and some green edit buttons on the top.

The size is also not setup automatically and it currently depends on the size of your browser window.

image

lpatiny:
We have currently 3 zones that should not be displayed

- the outline of the molecule
- 2 grey squares in the bottom

image

We after to find out how the NMRium div can be resized to either A4 or letter before printing.

lpatiny:
Google docs offers similar feature but in our case it will always be landscape:
image

lpatiny:
Looks great, thanks !

Couple of little points:

I found out that the place we put the delta(ppm) is different in 1D and 2D.

1D it is in the grey banner in the bottom:
image

2D it is over this grey banner
image

This has as consequence than when printing currently we need to display this banner and for 2D it appears like an empty grey banner:

image

2 possibilities:
- for 1D we move the d(ppm) over the grey bottom banner and this banner is hidden when printing
- for 2D we move the d(ppm) in the grey zone and when we print the grey banner becomes white

I would go for the first possibility even if we loose a little bit of space for the spectrum.


The second issue is how the size of the paper is expected to be decided ? I didn't see in the code the selection of the paper size and if I change to A3 for example the spectrum would not enlarge:



On firefox there are also 2 pages rather than one:

image

Did you know that you can emulate the @media: print from the developer tools?
SHIFT + CMD + P and then look for media. You can then inspect the dom of the page as it would be rendered for priting.

image

When I do this, we have the panels on the right and a grey zone at the bottom:

image

If you need a new feature to hide the panels when printing (from react-science), do not hesitate to ask us.

lpatiny:
@targos Hamed wrote me:

> I'm thinking, Instead of hiding the panels, toolbars, and header, and change the structure we can check whether it's possible to inject the viewer into an iframe and print from there. This needs to be investigated

You was speaking to me about some features of react that allows to copy dom possibly in another window. Could you detail a little bit this idea ?

targos:
My idea is really abstract for now. I don't know how to do it (the React feature is `createPortal`)

lpatiny:
Would using https://stackoverflow.com/questions/1234008/detecting-browser-print-event help in order to catch the 'File -> print' event ?

lpatiny:
In seems also that when you cancel the print dialog box there is no resize and we stay with the zoomed layout.

lpatiny:
Dialog box is really a great idea ! Could you save this information in the workspace as well so that it remembers the choice ? I don't think it is the case currently.

hamed-musallam:
it still not, We should save them within the workspace. Should we store them at the top level as follows?

```ts
{
  printPageOptions: {
    layout: 'portrait',
    size: 'A4',
    padding: '0cm'
  }
}
```
Or should they be included under the 'general' section in the 'General' tab of the general settings?

lpatiny:
As long as we always present the print dialog we don't need to add it in the general preferences but this could evolve in the future. It could become part of the general settings and be editable there as well.

lpatiny (approved):
Functionalities looks good to me.

hamed-musallam:
@lpatiny 

can you test again on Firefox, Chrome, and Safari

PR #2807, opened 2023-12-15
fix: resizing range to the same form,to should not add a new range

hamed-musallam:


src/data/data1d/Spectrum1D/ranges/changeRange.ts
```diff
@@ -28,7 +28,7 @@ export function changeRange(spectrum: Spectrum1D, range: Range) {
   ) as Signal1D[];
 
   if (index !== -1) {
-    spectrum.ranges.values[index] = mapRanges(
+    const [newRange] = mapRanges(
```
targos, line 31: @jobo322 Can you please have a look? Is it really expected that `mapRanges` returns an empty array if the range hasn't changed?

jobo322: mapRanges is used in several places in nmr-processing and nmrium. This filtering is useful in the case of auto ranges picking. For resizing, we should only pass the option `checkIsExisting` as false. So it will not try to filter the array.

targos: Can you please make a follow-up PR to fix this correctly?

Issue #2820, opened 2024-01-11
Problem with spectra having discontinuous domain


andcastillo:
I have some spectra in .nmrium format with discontinuous x. When I full zoom out, it fill the gap with some points around the hole. If you zoom in around the zone, it shows the right flat zone.

![case1](https://github.com/cheminfo/nmrium/assets/3810292/9ebd74f8-47c6-4659-b6b4-97af03e598f0)
![case2](https://github.com/cheminfo/nmrium/assets/3810292/b30d8b17-11c5-44d7-a0d3-27469bfa082c)
[toview.json](https://github.com/cheminfo/nmrium/files/13900949/toview.json)

Use the attached file to reproduce the problem. I changed the extension to .json.

lpatiny:
Indeed currently NMRium expect to have a constant deltaX because we are reducing the number of points in order to increase display speed.
So we need to discuss if this is an important feature and how to implement it.

lpatiny:
@andcastillo The culprit is likely https://github.com/mljs/spectra-processing/blob/14c0ddeecb8af069f98db44a54465c9f05edf5a2/src/xy/xyReduce.ts#L26 because we don't expect deltaX to change.

If you are able to improve the code without slowing it down you are welcome to make a PR.

jobo322:
what do you think about, split the discontinue spectrum into several spectrum, in that way nmrium could manage correctly those ROIs

PR #2825, opened 2024-01-12
fix: xy data with discontinuous x

andcastillo:
Solve the issue by getting the zones at before calling `xyReduce`.

Screenshot 2024-01-12 at 6 14 21 pm

Related to #2820


lpatiny:
After discussion we will check if the x values are ok in nmr-load-save and will currently not deal with non continuous X because other algorithms would crash or give unexpected results.

PR #2858, opened 2024-01-25
chore: fix label for paste molfile or SMILES

lpatiny:


lpatiny:
@hamed-musallam I corrected the label to paste molfile or SMILES but there is an issue in the prediction. When we paste a molecule the button 'Predict spectra' is not activated.

Please could you have a look ?

PR #2862, opened 2024-01-26
feat: invert actions

hamed-musallam:


lpatiny:
We need to add an exception because for the zoom button we should not invert the behaviour. It stays always the same even in the preference 'invert' is selected.

PR #2869, opened 2024-01-30
feat: show assignment atoms over ranges

hamed-musallam:


src/component/context/TopicMoleculeContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children,
+}: TopicMoleculeProviderProps) {
+  const moleculesRef = useRef<Record<string, TopicMolecule>>({});
+  const { molecules } = useChartData();
+
+  useEffect(() => {
+    for (const molecule of molecules) {
+      moleculesRef.current[molecule.id] = new TopicMolecule(
+        OCL.Molecule.fromMolfile(molecule.molfile),
+      );
```
lpatiny, line 33: If the TopicMolecule exists already we should update the TopicMolecule and not create a new one

```
const molecule = OCL.Molecule.fromMolfile(molfile);
existing.topicMolecule.fromMolecule(molecule);
```

PR #2879, opened 2024-02-01
refactor: resurrect spectrum from publication string

hamed-musallam:


michelgravel:
Multiplicities are not recognized properly.
For example, a 'd' shows up as a triplet, a 't' shows up as a quartet, etc.
This problem persists with more complex signals such as dd, dt, etc.

lpatiny:
@michelgravel Could you share with us the string you try to resurrect ? I just found out there is an issue with quint, hex, hept with could not find any issues with simple strings like:

`1H NMR (CDCl3, 400MHz) δ 1 (s, 1H), 2 (d, 1H, J=7), 3 (t, 1H, J=7), 4 (q, 1H, J=7), 8 (dd, 1H, J=7, J=4)`

https://github.com/cheminfo/nmrium/issues/3088

michelgravel:
It appears to be a browser issue.
It looks OK using Edge, but the problem arises when using Firefox.

lpatiny:
> It appears to be a browser issue. It looks OK using Edge, but the problem arises when using Firefox.

Thanks I would never have found this issue. Indeed the generated spectra are different but the ranges are identical. Very weird.

@targos I don't see what could be the origin of such a problem. We will have to check this together I think.

image

targos:
Please open a new issue. Discussions on closed items tend to be forgotten

Issue #2881, opened 2024-02-01
Fix .svg formatting of metadata for Word


rabatkri:
# Current Behaviour

It is possible to export the spectra as .svg file from NMRium. But there is a formatting problem with the metadata when importing the .svg into Microsoft Word.

When importing the .svg file into a Microsoft Word (tested with Version 16.81 MacBook Air 2017) the formatting of the metadata (top left corner) is wrong. For this a .pdf file obtained from Microsoft word is attached. 
When opening the same file in Google Chrome the formatting is not a problem or here as well. This problem also does not appear when using exported .png file. 

.svg file
![KRA-064-A_10](https://github.com/cheminfo/nmrium/assets/158492972/d021c848-784a-4950-af19-7e16bcf6f84c)

.png file
![KRA-064-A_10](https://github.com/cheminfo/nmrium/assets/158492972/c1d821ad-a503-4e63-96b5-a30106811b76)

# Obtained .pdf from Microsoft Word

[svg_vs_png_file.pdf](https://github.com/cheminfo/nmrium/files/14127737/svg_vs_png_file.pdf)

# Original data for possible reproduction

[KRA-064-A.zip](https://github.com/cheminfo/nmrium/files/14127753/KRA-064-A.zip)

Thanks! I'm happy to provide additional information.

lpatiny:
@rabatkri thanks for reporting this issue. We have identified the problem and it should be fixed in the next version.

You can already test it here: https://svg.nmrium.pages.dev/

rabatkri:
Thanks it works well and thanks for the quick fix @lpatiny

lpatiny:
@rabatkri the new version is now in production on www.nmrium.org

Could you register on https://www.nmrium.org/contact so that we can keep you informed about the new features ?

rabatkri:
@lpatiny thank you!

PR #2882, opened 2024-02-01
feat: improve SVG exportation

lpatiny:
closes: https://github.com/cheminfo/nmrium/issues/2881

### DIFF
--- a/src/component/1d/Chart1D.tsx
+++ b/src/component/1d/Chart1D.tsx
@@ -28,6 +28,7 @@ function Chart1D({ mode, width, height, margin, displayerKey }) {
   return (
     <PeakEditionProvider>
       <svg
+        style={{ fontFamily: 'Arial, Helvetica, sans-serif' }}
         viewBox={`0 0 ${width} ${height}`}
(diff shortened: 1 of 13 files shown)


lpatiny:
@targos I don't know why this test is failing now. The issue is there: https://github.com/cheminfo/nmrium/blob/5802722ceafb951bfd48f01f1ac12b7c8e566a9a/test-e2e/panels/peaks.test.ts#L74-L82

targos:
It's not obvious, the error is:
```
<svg width="701" id="nmrSVG" height="786" viewBox="0 …>…</svg> intercepts pointer events
```

targos:
I personally can reproduce the same errors outside of this PR (main branch), but on GitHub actions, it is specific to this PR (https://github.com/cheminfo/nmrium/pull/2884 passed for example)

targos:
Somehow today I couldn't reproduce the problem on main. Found that the issue was related to the font change. We need to ask PW to hover the text element, otherwise it can fall between the text and the annotation, which doesn't trigger the hover.

PR #2895, opened 2024-02-06
fix: correct rollback spectrum behaviour

### LINKED ISSUE #2876: Change tool from baseline does not cancel the correction
If you select the baseline correction tool and then select another tool the preview of the baseline correction is not cancelled.

### LINKED ISSUE #2889: Baseline correction strange behaviour
If I click twice on the `Baseline correction` button the spectrum changes. This seems strange.

Also if I cancel the baseline correction I have a vertical rescale.

hamed-musallam:
close #2876
close #2889


lpatiny:
It works so much better !

PR #2896, opened 2024-02-07
feat: resize baseline zones

hamed-musallam:


lpatiny:
Works as expected. Thanks.

PR #2898, opened 2024-02-08
fix: spectra preferences validation

hamed-musallam:


lpatiny:
If I empty completely the field I still have the error

image

PR #2920, opened 2024-02-21
fix: experiment type by pulse sequence

jobo322:


lpatiny (approved):
Functionality looks ok to me.

PR #2922, opened 2024-02-22
refactor: panels toolbar 

hamed-musallam:


src/component/utility/toString.ts
```diff
@@ -0,0 +1,10 @@
+interface ToStringOptions {
+  trueLabel?: string;
+  falseLabel?: string;
+  negate?: boolean;
```
targos, line 4:
```suggestion
```

Unused option.

src/component/utility/toString.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export function toString(flag: boolean, options: ToStringOptions = {}) {
```
targos, line 7: Try to find a better name because this is way too generic and too easy to confuse with the builtin JS method.

PR #2949, opened 2024-03-06
2933 improve peak picking

jobo322:


jobo322:
@hamed-musallam @lpatiny 
With the last version of ml-spectra-processing appear the incompatibility with [`IntegralData`](url)

```
interface IntegralData {
  x: Float64Array;
  y: Float64Array;
  // eslint-disable-next-line react/no-unused-prop-types
  id: string;
}
```
The function `xyIntegral` returns DataXY<number[]>. 

One line that has the issue is:
https://github.com/cheminfo/nmrium/blob/2933-improve-peak-picking/src/component/1d/ranges/RangesIntegrals.tsx#L83

Should I refactor the interface `IntegralData`?

jobo322:
@hamed-musallam @lpatiny 
another incompatibility appear in 

```
const matrix = matrixZPivotRescale(
    isFid
      ? (spectrum.data as NmrData2DFid).re.z
      : (spectrum.data as NmrData2DFt).rr.z,
    {
      max: 255,
      ArrayConstructor: Int16Array,
    },
  );
  ```

in the line: 
https://github.com/cheminfo/nmrium/blob/2933-improve-peak-picking/src/component/2d/fid/FidCanvas.tsx#L119-L127

with log error: 
`Type 'Int16ArrayConstructor' is not assignable to type 'DoubleArrayConstructor | undefined'`

**Comment by lpatiny** [comment]
Concerning https://github.com/mljs/spectra-processing/blob/main/src/xy/xyIntegral.ts I think we should change this code and return a Float64Array. We can guess correctly the length of the array so we can create it. Let's call this a 'fix' and not a breaking change ...

**Comment by lpatiny** [comment]
Please also rebase the project.

**Comment by lpatiny** [comment]
It looks much better !

When we do auto range picking we have 's' while when we resize we have 's br'. Could we also get the broad information directly ?

image

**Comment by lpatiny** [comment]
The range picking is worse now with Cytisine. What used to be

image

Becomes a big range containing multiplets:

image

File: package.json [hunk 1]
```diff
@@ -150,4 +150,4 @@
     "vite": "^5.1.3",
     "vitest": "^1.3.1"
   }
-}
+}
```
targos: Can you revert this? And maybe check why this whitespace is removed by your editor.

PR #2953, opened 2024-03-08
feat: edit range peaks

hamed-musallam:


lpatiny:
I added peaks manually but they don't appear in the spectrum even if I toggle the show peaks button.

image

When we click on a place we should keep all the digits of the chemical shift. It is only when we display that we should round the number. But this could be done later, not that important.

PR #2964, opened 2024-03-14
feat: zoom in/out with the mouse wheel

### LINKED ISSUE #2962: Zooming in and out with the mouse wheel
I suggest adding a functionality to zoom in and out of the spectrum in a gradual fashion rather than by selecting a zone. Some commercial software have this functionality and it is very useful when we want to increase or decrease the zoom window. Less annoying than double clicking to a previous zoom window. For example, one could hold the right mouse button and scroll the mouse wheel at the same time.

hamed-musallam:
close #2962

### DIFF
--- a/src/component/1d/Viewer1D.tsx
+++ b/src/component/1d/Viewer1D.tsx
@@ -293,8 +293,8 @@ function Viewer1D({ emptyText = undefined }: Viewer1DProps) {
   }, [dispatch]);
 
   const handleZoom = useCallback<OnZoom>(
-    (event) => {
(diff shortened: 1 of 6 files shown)


lpatiny:
Works very well ! Maybe we could open another issue to implement something similar for 2D spectra.

lpatiny:
@targos Do you have an idea why we the action Cloudflare Pages fails since couple of days ? The status can not be reported.

targos:
It seems that Cloudflare changed the name of the check. I updated the requirements to use the new one.

lpatiny:
Great ! Works very well.

lpatiny (approved):

PR #2970, opened 2024-03-18
fix: improve consistency of labels

targos:
- Do not use uppercase in wrong places.
- Remove unnecessary whitespaces.
- Refactor Label component to wrap the form element with `<label>` to
  remove the need for `htmlFor` and automatically bind it to the element.
- Drive-by: correct spelling error (spectrums instead of spectra).


hamed-musallam:
I believe it's better to include a space before and after the shortcut label and capitalize the letter 'P' in 'Press'.

image

targos:
I can bring back the capital P, but in English there are no spaces inside paretheses.

hamed-musallam:
> I can bring back the capital P, but in English there are no spaces inside parentheses.

Yes, you're correct, but it was more readable than having no spaces. You can skip my comment, it is not important and you can keep it like that

hamed-musallam (approved):

PR #2971, opened 2024-03-19
feat: create and edit assignment labels for ranges

hamed-musallam:


hamed-musallam:
@lpatiny 

Would you like to add a new property to the range preferences for toggling the visibility of the assignment label column? If so, what would be the default value?

We currently have one property to toggle the visibility of the range and signal assignment columns. I adjust the visibility of the assignment label based on this property as well

Screenshot 2024-03-19 at 11 52 35

hamed-musallam:
f you require a new property to toggle the visibility, I have created a PR in nmr-load-save for that: https://github.com/cheminfo/nmr-load-save/pull/367. Otherwise, I will close this PR.

What do you think? Is a new property necessary, or is what we currently have sufficient?

lpatiny:
I think it should not be visible by default.
I would also prefer it is the second column just after # and before delta.
Is there a way to display the assignment on the spectrum ? This would be very practical. We will have conflict between labels that overlaps but we will solve this later. It could be on the same line as the 12 of this image. This would also require one more toggle in the range header ...
image

lpatiny:
Please also move this icon

image

just after the sum symbol because there are 'linked'.

PR #2973, opened 2024-03-19
fix: set the size correctly for real part in slicing function

hamed-musallam:
We have encountered a major bug that causes failures in the slicing tool and 2D phase correction tool. This issue was introduced by the PR 
- https://github.com/cheminfo/nmrium/pull/2949

### DIFF
--- a/src/data/data2d/Spectrum2D/getSlice.ts
+++ b/src/data/data2d/Spectrum2D/getSlice.ts
@@ -104,7 +104,7 @@ export function getSlice(
 function initiateData(from: number, to: number, size: number): NmrData1D {
   return {
     x: xSequentialFillFromTo({ from, to, size }),
-    re: new Float64Array(length),
+    re: new Float64Array(size),
   };
 }
 
--- a/test-e2e/tools/slicing.test.ts
+++ b/test-e2e/tools/slicing.test.ts
@@ -1,0 +1,28 @@
+import { test, expect } from '@playwright/test';
+
+import NmriumPage from '../NmriumPage';
+
+test('Slicing too should show 1d (horizontal and vertical) traces', async ({
+  page,
+}) => {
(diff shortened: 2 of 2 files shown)


src/data/data2d/Spectrum2D/getSlice.ts
```diff
@@ -104,7 +104,7 @@ export function getSlice(
 function initiateData(from: number, to: number, size: number): NmrData1D {
   return {
     x: xSequentialFillFromTo({ from, to, size }),
-    re: new Float64Array(length),
+    re: new Float64Array(size),
```
targos, line 107: How was `length` not reported as non-existing variable?

hamed-musallam: it is strange !!!!  the `"strict": true` is in the tsconfig` but it is not reported

targos: Found it: it's one of the many available globals on `window`:
image

I'll add it to our [list of restricted globals](https://github.com/cheminfo/eslint-config-cheminfo-react/blob/main/base.js#L24-L31)

hamed-musallam: It is possible that length is defined as a global variable somewhere, but I've tried searching for it and couldn't find where it's defined

hamed-musallam:
@lpatiny

After merging this PR, we'll need to create a new release of NMRium

src/data/data2d/Spectrum2D/getSlice.ts
hamed-musallam: Alright, I'll mark it as resolved and merge this PR. Updating ESLint rules can be done in another PR. What do you think?

targos: Yes, of course

PR #2984, opened 2024-03-22
fix: spectra simulation options validation

hamed-musallam:


src/component/panels/spectrumSimulation/SpectrumSimulationSimpleOptions.tsx
```diff
@@ -58,6 +58,7 @@ export default function SpectrumSimulationSimpleOptions({
               textAlign: 'center',
             },
           }}
+          step={0.1}
```
targos, line 61: You can also add `min={0.1}` to prevent the stepper from going into invalid values.

PR #2985, opened 2024-03-22
fix: re-initialize FifoLogger instance when opening the modal

### LINKED ISSUE #2981: Clear log when generating spectrum
If you generate a spectrum from a string the logs is not cleared when closing the window

image

This means that if you open again the dialog you see the previous logs and this should not be the case.

hamed-musallam:
close #2981


lpatiny:
It fix the issue. Thanks!

PR #3008, opened 2024-04-04
feat: improve main toolbar tooltip

hamed-musallam:
It's a sample. We'll discuss what to set there later.

Screenshot 2024-04-04 at 15 19 58
Screenshot 2024-04-04 at 15 20 06


lpatiny:
Because the shortcuts only work with lowercase I think they should also be indicated there as lowercase

image

For zoon help: 3 possibilities (not 2):
- Vertical zoom: scroll wheel
- Horizontal zoom: shift + scroll wheel
- Pan: right click

 Remove 'Learn more' if there is currently no links to docs.nmrium.org

PR #3009, opened 2024-04-04
feat: do not split molfile containing many molecules

lpatiny:
Problem is that there is no way to draw a salt and this is quite annoying. People should take care to draw many molecules if needed.


lpatiny:
Need to be done in a better way. Opening an issue: https://github.com/cheminfo/nmrium/issues/3087

PR #3018, opened 2024-04-09
feat: zone assignment label

hamed-musallam:
https://github.com/cheminfo/nmrium/assets/35760236/6a8d5c02-551c-49ad-b4f8-4c4435e808e9


lpatiny:
Great functionality and for me it can be merged !

Now I could still see big improvement if we can directly add and edit the assignment from the spectrum when the **zone tool is selected**.

If you SHIFT + click (or just click ?) on a zone:
- if no assigment is present we add one add we go to edit mode
- if there is an assignment it is placed in edition mode and the full text is selected

If you click on a label it goes to edition mode.

We will also have to think about the label size (they may have to be relative size rather than absolute) in the future but this I will do assigning cyclosporin.

PR #3020, opened 2024-04-09
fix: create a simple signal if data length is smaller than 7 points

jobo322:


lpatiny (approved):

PR #3032, opened 2024-04-26
feat: zoom out based on defined axis domain preferences

### LINKED ISSUE #3024: Hardcoded axis values
I had a request to be able to manually enter the values from/to of the axis in order to consistently generate spectra with the same scale. This makes sense, but I don't know how to do this in an intuitive way so that if we set the values and zoom 'full out', we don't loose those settings.

We could consider the possibility in the General preferences to have a predefined 'from' / 'to' for each nucleus, but I don't know how we could apply / recall those settings when manipulating the spectrum. Maybe it could be an alternate shortcut of the full out like SHIFT + click or SHIFT + 'f'.

hamed-musallam:
Define the axis domain (x-axis for one-dimension spectra and x-axis, y-axis for two-dimension spectra) from the spectra panel preferences. Once it is defined, use Shift + f to zoom out.

close #3024


lpatiny (changes requested):
Currently if you set a value to 0 (for 1D or 2D), it is not taken into account 

image

hamed-musallam:
@lpatiny 

I am still convinced that the domain preferences should be in the general settings and not in the spectra panel preferences, the panel preferences are only for the panel itself and not a general setting that is applied over the displayer. I think it could be somewhere in the general settings under the General tab or we could have a separate tab

image

PR #3034, opened 2024-05-02
feat: improve zone assignment label addition and editing functionality

hamed-musallam:


lpatiny:
Could you change the highlight behaviour for the text ? Currently when highlighted it is less visible. I would suggest to simply but the text in bold and not change the color (or use dark red as color).

image

PR #3038, opened 2024-05-03
feat: improve range assignment label addition and editing functionality

hamed-musallam:


hamed-musallam:
Hello @targos, Can you give me a hand in this issue, The vendor chunk exceeds 25 MB, preventing Cloudflare from building and publishing the project. Do you have any suggestions? Here's what we currently have in the vendor chunk

image

lpatiny:
I think you can just a manual chunk in vite.config.mts for blueprintjs
```
            if (id.includes('node_modules/openchemlib/')) {
              return 'openchemlib';
            }
            if (id.includes('node_modules/@blueprintjs/')) {
              return 'blueprintjs';
            }
            if (id.includes('node_modules')) {
              return 'vendor';
            }
```

lpatiny:
Building with `sourcemap: 'inline', ` rather than `sourcemap: true,` would help but I don't know if this is acceptable.

targos:
> Building with `sourcemap: 'inline', ` rather than `sourcemap: true,` would help but I don't know if this is acceptable.

Why would it help? I expect it to be worse: more bytes in the same files + base64-encoding.

targos:
I suggest to split the vendor chunk in more files like we did with openchemlib.

hamed-musallam:
The size was reduced by over 50% after switching to inline source maps:

`sourcemap =true`
Screenshot 2024-05-08 at 09 25 02

`sourcemap ='inline'`
Screenshot 2024-05-08 at 09 27 33

hamed-musallam:
Indeed, the size of the JavaScript file has increased, but we avoid large files exceeding 25MB

targos:
Please verify that the debugging experience with devtools is still good on dev.nmrium.org. If that's the case I'm fine with the change because it only affects development and test builds.

hamed-musallam:
the size is reduced by more than  half once I set the source map inline

> dev.nmrium.org

Let's stick with this approach for now, and we can consider chunking the vendor JavaScript in the future

PR #3050, opened 2024-05-17
fix: do not calculate noise every time the wheel is actioned

jobo322:


lpatiny (changes requested):
The functionalities looks better but the test are not passing. Please have a look why.

@hamed-musallam Please review the code as well.

src/component/reducer/actions/ToolsActions.ts
```diff
@@ -726,9 +726,9 @@ function levelChangeHandler(draft: Draft<State>, action: LevelChangeAction) {
 
   try {
     for (const spectrum of spectra as Spectrum2D[]) {
-      const contourOptions = spectrum.display.contourOptions;
-      const zoom = contoursManager(spectrum.id, levels, contourOptions);
+      const zoom = contoursManager(spectrum, levels);
       levels[spectrum.id] = zoom.wheel(deltaY, altKey);
+      levels[spectrum.id] = zoom.checkLevel();
```
targos, line 731: Two lines that assign to the same value means either there's a bug or the first line is useless.

src/component/2d/ft/Contours.tsx
```diff
@@ -129,28 +132,33 @@ function ContoursInner({ spectra, displayerKey }: ContoursInnerProps) {
...
-              sign="negative"
-              spectrum={spectrum}
-              color={spectrum.display.negativeColor}
-              onTimeout={timeoutHandler}
-            />
-          )}
-        </g>
-      ))}
+      {spectra?.map((spectrum) => {
+        const noise = calculateSanPlot('2D', (spectrum.data as NmrData2DFt).rr);
```
targos, line 136: @lpatiny Does it make sense to calculate the noise everytime a spectrum is rendered?

src/data/data2d/Spectrum2D/contours.ts
```diff
@@ -76,72 +101,88 @@ function contoursManager(
...
+      (minNegativeLevel > 0 && sign === -1) ||
+      (minNegativeLevel <= maxNegativeLevel - negative.numberOfLayers &&
+        sign === 1)
     ) {
-      currentLevel.negative += sign;
+      contourOptions.negative.contourLevels[0] += sign * 2;
     }
   }
+  contourOptions.negative.contourLevels[0] =
+    contourOptions.negative.contourLevels[0];
```
targos, line 142: You're assigning a value to itself

lpatiny:
@hamed-musallam Could you change the behaviour so that when we move the sliders the spectrum is updated on the fly (and not on mouse release like currently) ?

A small debounce will certainly be required.

targos:
Do we want these new very large files in the repo?

lpatiny:
> Do we want these new very large files in the repo?

@jobo322 Please use 512 x 512, no need to have 1024 on 1024 I think.
Also don't do it 'pretty'. Simply JSON.stringify without the '2'

The files should then be around 1.3 Mb. Just by removing the \n + spaces I went from 22Mb to 5.5 Mb.

lpatiny:
Functionality looks ok to me but for the manual change of slider that should update the contour plot after debouncing (before release of the cursor).
@hamed-musallam Could you check this ?

hamed-musallam:
What is left is to refactor the spectrum object and move the contours Options object to `view` which can be done in another PR


``` ts

{
contourOptions:{
          positive: {
            contourLevels: [15, 100],
            numberOfLayers: 10,
            numberOfZoomLevels: 10,
          },
          negative: {
            contourLevels: [15, 100],
            numberOfLayers: 10,
            numberOfZoomLevels: 10,
          }
}
```

hamed-musallam (approved):

hamed-musallam:
> The functionalities looks better but the test are not passing. Please have a look why.
> 
> @hamed-musallam Please review the code as well.

it is fixed

lpatiny (approved):
Functionality looks ok to me

src/data/data2d/Spectrum2D/contours.ts
```diff
@@ -43,105 +44,127 @@ interface ReturnContoursManager {
...
+    Math.abs(quadrantData.minZ),
+    Math.abs(quadrantData.maxZ),
+  );
+
+  const minLevel = Math.ceil(
+    10 *
+      Math.log2(
+        1 + (3 * xMaxAbsoluteValue([positive, negative]) * (2 ** 10 - 1)) / max,
+      ),
+  );
```
targos, line 67: This calculation needs a comment or to be moved to a function that has a descriptive name and jsdoc.

src/data/data2d/Spectrum2D/contours.ts
```diff
@@ -43,105 +44,127 @@ interface ReturnContoursManager {
...
-    for (let i = 1; i <= length; i++) {
+    for (let i = 1; i < length + 1; i++) {
       factors[i] = factors[i - 1] + (exp - 1) / exp ** i;
     }
-
-    const lastFactor = factors[length];
-
-    const result = new Array(length);
-
+    const lastFactor = factors[length - 1];
```
targos, line 160: This is not the last factor.

src/data/data2d/Spectrum2D/contours.ts
```diff
@@ -157,66 +180,62 @@ function range(from: number, to: number, step: number) {
...
-
   const ys = getRange(data.minY, data.maxY, data.z.length);
   const conrec = new Conrec(data.z, { xs, ys, swapAxes: false });
-  const sanResult = calculateSanPlot('2D', data);
-  const median = sanResult.positive;
-
-  const max = Math.max(Math.abs(data.maxZ), Math.abs(data.minZ));
+  const max = Math.max(Math.abs(data.minZ), Math.abs(data.maxZ));
+  const minLevel = (max * (2 ** (boundary[0] / 10) - 1)) / (2 ** 10 - 1);
+  const maxLevel = (max * (2 ** (boundary[1] / 10) - 1)) / (2 ** 10 - 1);
```
targos, line 219: Same here, calculation is not obvious.

targos:
Can be merged after addressing my comments.

src/data/data2d/Spectrum2D/contours.ts
jobo322: it is the inverse function of max * (2 ** (level / 10) - 1)) / (2 ** 10 - 1).

hamed-musallam:
Can we merge this PR, or are there still comments?

PR #3055, opened 2024-05-21
feat: generate and display  stocsy

hamed-musallam:


targos:
cf build error is:
```
15:18:52.548 | x Build failed in 6.21s
-- | --
15:18:52.550 | error during build:
15:18:52.550 | RollupError: Could not resolve "../../context/ExportPrepareContext" from "src/component/1d/matrix/Stocsy.tsx"
15:18:52.550 | file: /opt/buildhome/repo/src/component/1d/matrix/Stocsy.tsx
15:18:52.550 | at getRollupError (file:///opt/buildhome/repo/node_modules/rollup/dist/es/shared/parseAst.js:376:41)
15:18:52.551 | at error (file:///opt/buildhome/repo/node_modules/rollup/dist/es/shared/parseAst.js:372:42)
15:18:52.551 | at ModuleLoader.handleInvalidResolvedId (file:///opt/buildhome/repo/node_modules/rollup/dist/es/shared/node-entry.js:18891:24)
15:18:52.551 | at file:///opt/buildhome/repo/node_modules/rollup/dist/es/shared/node-entry.js:18851:26
15:18:52.629 | Failed: Error while executing user command. Exited with error code: 1
```

lpatiny:
1. Please use SVG and allow vertical scale. We will improve with grouped path by color very soon like Norman did in the visualizer and JsGraph
2. For now please use the first row as Y axis and paint the color with the returned value from stocsy
3. deal with the index on which the user clicked

lpatiny:
If I only display the STOCSY

image


And I just moved the scroll wheel (without ALT), nothing happened (as expected), but nevertheless, the STOCSY was redrawn.

image

lpatiny:
@hamed-musallam For the boxPlot and the calculation of the 3 paths please use `xyReduce`. Becuase we have only color it can be used in this case.

lpatiny:
It seems that currently the SVG is painted outside the visible div as well. This can be observed when you donwload the SVG and open it in inkscape

image

And this could explain it is also relatively slow when zoomed in.

src/component/1d/matrix/Stocsy.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return <InnerStocsy scaleRatio={scaleRatio} chemicalShift={chemicalShift} />;
+}
+export function InnerStocsy({ scaleRatio, chemicalShift }) {
+  const matrix = useMatrix();
+
+  if (!matrix) return null;
+
+  const { x, matrixY } = matrix;
+  const xIndex = xFindClosestIndex(x, chemicalShift ?? x[0], { sorted: false });
+  const { color, y } = matrixToStocsy(matrixY, xIndex);
```
lpatiny, line 48: This method, matrixToStocsy, takes 30% of the time and is recalculated currently continuously. This should not be the case if matrixY and xIndex does not change.

src/component/1d/matrix/Stocsy.tsx
```diff
@@ -0,0 +1,159 @@
...
+    xDomain: [from, to],
+  } = useChartData();
+
+  return useMemo(() => {
+    if (!matrix) return null;
+
+    const { x, matrixY } = matrix;
+
+    const cIndex = xFindClosestIndex(x, chemicalShift ?? x[0]);
+    const { color, y } = matrixToStocsy(matrixY, cIndex);
```
lpatiny, line 43: matrixToStocsy should not be called when we zoom and I don't think the Memo should be triggered by from and to.

src/component/1d/matrix/useMatrix.ts
```diff
@@ -0,0 +1,153 @@
...
+      if (x[i] >= to) {
+        toIndex = i;
+        break;
+      }
+    }
+  }
+  return { fromIndex, toIndex };
+}
+
+export function sliceArray(
```
lpatiny, line 40: It is not clear to me why this method exists and we need to keep the first and last point. Should we not just keep the point just before and after (if they exists !) ? If really required it should be renamed to something like 'sliceArrayAndKeepFirstLast'

Also at the implementation point of view doing an Array.slice will likely be much faster. At worse you could slice 2 points more and change the first and the last one.

And concerning types you call this method with color (an array of string) and NumberArray includes already number[] and Float64Array.

src/component/1d/matrix/useMatrix.ts
```diff
@@ -0,0 +1,153 @@
...
+  inputRange: [number, number],
+  outputRange: [number, number],
+) {
+  return (value) =>
+    outputRange[0] +
+    ((value - inputRange[0]) * (outputRange[1] - outputRange[0])) /
+      (inputRange[1] - inputRange[0]);
+}
+
+export function mapMatrixColors(colors) {
```
lpatiny, line 69: I don't think this function is used and could be deleted

src/component/1d/matrix/useMatrix.ts
```diff
@@ -0,0 +1,153 @@
...
+  result[0] = array[0];
+  for (let i = 0; i < toIndex - fromIndex; i++) {
+    result[i + 1] = array[fromIndex + i];
+  }
+  result[resultLength - 1] = array.at(-1);
+
+  return result;
+}
+
+export function interpolateNumber(
```
lpatiny, line 59: I don't think this function is used and could be removed

lpatiny:
Still small improvements that could be done:

1. Hiding spectra yield to an error in the console



2. Confusion with 'Apply processing' and 'Update processing'

The first time we open Matrix generation the button should be labeled 'Apply processing' and after we could have 2 buttons:
- Update processing: if any of the parameters change and there is a reason to update the processing
- Remove processing: this will remove the filter 'Signal processing' from all the spectra. This button should be active if there is at least one spectrum with 'Signal processing'

3. Show stocsy / boxplot

The 2 toggle button should only be available if the matrix has been generated and therefore that 'Apply processing' was clicked. We should not have a matrix before the 'Apply processing' because if we superiimpose spectra from different origins we will be in trouble.

So those button should rely on the presence of the matrix.

4. Vertical scroll generates a lot of internal work by react (I guess)

Not sure how this can be improved but vertical scroll wheel even if there is no redraw takes around 100mS

hamed-musallam:
> 1. Hiding spectra yield to an error in the console

The warning you see is from Blueprint.js, indicating that a tooltip is active but its content is either empty, null, or undefined.

I will implement a temporary fix for this issue in NMRium. However, the underlying problem should be addressed in the react-science button component by checking the tooltip content and disabling it if it is empty.

PR #3084, opened 2024-06-07
feat: change default sum value from 100 to 1

### LINKED ISSUE #3056: Ranges / integrals, default behaviour and value
The current behaviour of the range / integral tool with a sum value and relative integration is somewhat confusing to synthetic chemists. I suggest that the first integral placed should always have an absolute value of 1.

@evaziegler

hamed-musallam:


lpatiny (changes requested):
The sum should stay 100 by default. This is what we use for structural analysis exercises.

The change that should be done is the possibility to have the link disabled based on workspace preferences.

image

And in this case, the relativeH of the first drawn integral is 1.

So what appears in the following video as 282000 should be 1.



The tip of the button could also be renamed (maybe move to the new  tip component)

image

Should become:

'Fixed sum of integrations'

PR #3148, opened 2024-07-22
Migrate ConfirmDialog

wadjih-bencheikh18:
https://github.com/cheminfo/nmrium/issues/1537


wadjih-bencheikh18:
I could't find a clean higher-level changes so I started with basic migration of dialogConfirm modals and others models will be in another PR

hamed-musallam:
@wadjih-bencheikh18   I was working on refactoring the confirmation dialog and have created a new component called Alert to handle this in NMRium

PR #3153, opened 2024-07-23
baseline corrections and precalculated zone

jobo322:


lpatiny:
If I test on the 13C without line broadeining it crash

image

image

hamed-musallam:
The real part values are undefined 
image

PR #3154, opened 2024-07-23
fix: import from publication string with common names of multiplicities

jobo322:
https://3088-import-multiplet.nmrium.pages.dev/


lpatiny (approved):
Functionality LGTM

PR #3159, opened 2024-07-25
feat: improve selection of spectra

### LINKED ISSUE #3158: Improve selection of spectra
Sometimes we need to hide all spectra but the one selected. We could have more button in the header of the Spectra panel:

image

Maybe in an 'Advanced' mode in the general preferences ?

Currently we have 2 buttons:
- hide selected spectra
- show selected spectra

The goal is to have 5

image

- show all spectra (this icon exists in the matrix panel)
- add selected spectra to shown spectra (this icon exists in the spectra panel)
- focus on current selected spectra (all spectra becomes invisible but the currently selected one)
- hide selected spectra from the currently shown spectra (this icon exists in the spectra panel)
- hide all the specvtra (this icon exist in the matrix panel)

There should be 5 different icons and the same should be used in spectra (all the 5 will be available) and in matrix where only 2 are available (for now)

Because it is not that obivous how to use it please plan an advanced tooltip so that I can write more help.

hamed-musallam:
close #3158

### DIFF
--- a/src/component/hooks/useToggleSpectraVisibility.tsx
+++ b/src/component/hooks/useToggleSpectraVisibility.tsx
@@ -1,9 +1,30 @@
-import { FaEye, FaEyeSlash } from 'react-icons/fa';
+import { FaEye, FaEyeSlash, FaMinus, FaPlus } from 'react-icons/fa';
+import { GrRadialSelected } from 'react-icons/gr';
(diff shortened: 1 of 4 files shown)


lpatiny:
In order to keep the interface simple please add the 3 new buttons:

image

Only if we have more or equal to 5 spectra. It does not really make sense if we have few spectra.

PR #3162, opened 2024-07-26
feat: export as JCAMP-DX includes all metadata

jobo322:


lpatiny:
@jobo322 I think you should revert this package-lock change

lpatiny:
We should not have this checkbox to save only real but the drop-down menu should depends on the files we have.

image

If original is FID we should add in the menu:
- Original FID

If original is FT we should add in the menu depending if we are real or real + imaginary
- Original real
- Original real + imaginary

The same is true for processed depending if it is FID, real or real + imaginary.

lpatiny:
If on indometacin I do a processing (baseline correction) I would expect to see 4 items in the menu (For original and processed) but currently it is not the case.

image

Otherwise it looks great. Thanks !

lpatiny:
When exporting processed data we still need to check if it is still a FID or after FT.

So in the case of the FID the message should be Processed FID (always real + imaginary)

image

PR #3166, opened 2024-07-29
fix: extract the atoms correctly

hamed-musallam:


lpatiny:
Could you add the link to the issue ?

lpatiny (approved):

PR #3174, opened 2024-08-01
feat: avoid splitting molecule fragments into separate molecules

hamed-musallam:


lpatiny (approved):
Functionality looks good to me. Thanks!

PR #3193, opened 2024-08-12
refactor: handle past mol or SMILES error in the main logger

hamed-musallam:


targos (dismissed):
There are a few occurrence of the typo: past -> paste

I don't think this makes sense to be handled at the level of the logger. These error should never reach the logger in the first place.

PR #3199, opened 2024-08-15
fix: keep matrix range unchanged when adding exclusion zones

### LINKED ISSUE #3198: Exclusion zone selection
When you want to select an exclusion zone in matrix and you zoom in during the process, the from / to values will also change

hamed-musallam:
close #3198


lpatiny (approved):
Functionalities LGTM

PR #3203, opened 2024-08-16
fix: restore pivot functionality and add test case

hamed-musallam:


lpatiny:
Great, thanks. Could you change the default color value? Because by default, it is red, and the spectrum as well.
You may put it in yellow, like for the baseline.

PR #3204, opened 2024-08-16
fix: automatic phase correction if phase correction values is undefined

jobo322:


lpatiny:
But this file still does not work seems to me.

[Resveratrol1_11280u600u_MeOD_08122024_clw_qH_ns64-1-1.jdf.zip](https://github.com/user-attachments/files/16664976/Resveratrol1_11280u600u_MeOD_08122024_clw_qH_ns64-1-1.jdf.zip)

PR #3210, opened 2024-08-23
refactor: panels preferences

hamed-musallam:
Use React Hook Form instead of Formik in panels preferences

### DIFF
--- a/src/component/panels/IntegralsPanel/IntegralsPreferences.tsx
+++ b/src/component/panels/IntegralsPanel/IntegralsPreferences.tsx
@@ -1,18 +1,11 @@
-import { Formik } from 'formik';
-import {
-  useCallback,
-  useImperativeHandle,
-  useRef,
-  memo,
-  forwardRef,
-  useMemo,
-} from 'react';
+import { useCallback, memo, forwardRef, useMemo } from 'react';
+import { useForm } from 'react-hook-form';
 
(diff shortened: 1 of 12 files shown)


src/component/elements/FomratField.tsx
targos, line 1: Fomrat -> Format

src/component/panels/IntegralsPanel/IntegralsPreferences.tsx
```diff
@@ -84,63 +79,51 @@ function IntegralsPreferences(props, ref) {
...
+  const { handleSubmit, control } = useForm<any>({
+    defaultValues: preferencesByNuclei,
+  });
 
-  useImperativeHandle(ref, () => ({
-    saveSetting: () => {
-      formRef.current.submitForm();
-    },
-  }));
+  useSettingImperativeHandle(ref, handleSubmit, saveHandler);
```
targos, line 86: You probably don't need to do this. If it's because the submit button is not inside the form, there is a way to link it using an id: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form

hamed-musallam: The button being outside the form is one reason, but I also need to check the validation of the form before closing the preferences panel

PR #3230, opened 2024-09-11
feat: customize assignment highlight color

hamed-musallam:


hamed-musallam:
- https://github.com/zakodium-oss/react-science/pull/758

PR #3237, opened 2024-09-12
feat: PNG export resolution

hamed-musallam:


targos:
Having the concept of a "resolution" (in DPI) in this context really feels weird to me, because a spectrum has no size in the real world, but ok...

hamed-musallam:
Yes, I'm aware of this 💯, but this is what came to mind as it seems close to what we're aiming for. It's simply a matter of rescaling the spectrum to meet the desired needs.

lpatiny:
The functionality looks ok to me. In the future probably we will indeed need to specify the size of the image like we do for printing to PDF. And then dpi would indeed make sense. Did well to plan an object for the options ;)

PR #3238, opened 2024-09-12
feat: display tracking for selected spectra

### LINKED ISSUE #3231: Tracking should display only the selected spectra
In tracking mode of multiple spectra analysis it would be better to display only the currently selected spectra.

This would allow to cherry pick some spectra and be able to find their corresponding reference.

image

hamed-musallam:
When one or more spectra are selected, display tracking for those specific spectra. If no spectra are selected, display tracking for all available spectra

close #3231


lpatiny:
Functionality looks ok to me.

PR #3242, opened 2024-09-13
fix: ensure shortcuts are not ignored when the focused element is a r…

hamed-musallam:
…adio or checkbox


lpatiny:
Functionality LGTM

PR #3243, opened 2024-09-13
wip: testing conrect

lpatiny:


lpatiny:
@hamed-musallam This is the POC of the cache. I just quickly did it as a global variable but it should of course be done at the level of each spectra using I guess a useRef.

I also did a small fix in ml-conrec to improve performances when nothing should be calculated.
@hamed-musallam Could you take over this PR or use it in the other PR ?
You can also be interesping to keep an eye on the memory using a logger.trace info with the total sum of all the lines in the cache.

Once done we will ask Michael to check if it is ok.

lpatiny:
If a check the size of the cache
```
  for (const level of _range) {
    console.log(level, cache.has(level), cache.size);
    if (cache.has(level)) {
```
I read 414. There could be an issue with rounding calculating the different levels ? I think calculate the possible levels should be done once for ever per spectrum to avoid this issue.

This was done with 2D cosy
image

image

lpatiny:
@targos Here are some comments / ideas.

If both positive and negative curves need to be calculated we should think about doing both at once. It seems much faster.

In conrec it would be better to have a limit in the number of segment per layer. Currently we have only a global timeout and if this timeout is reached nothing will be displayed. This could also prevent some crash with out of memory despite the fact that we are using typed array in order to limit this risk.

We could think about having fifo-logger sent to conrec to get some debug information as well as the error if the timeout was reached.

Conrec could work in a worker. There are only 101 levels possible for positive and negative. We could start calculations for the 10 default levels and possibly start in a webworker the calculation of the other 91 levels. Because we are using typedarray the transfer of the data should be fast.

PR #3248, opened 2024-09-17
feat: improve filters panel

hamed-musallam:


lpatiny:
The functionnality looks very good ! Thanks @hamed-musallam !

PR #3252, opened 2024-09-27
feat: improve save as PNG, SVG and copy to clipboard

hamed-musallam:
Browsers have limits on the maximum size of a canvas, These limits differ by browser, Chrome's maximum canvas size is 16,384 x 16,384 pixels.

Chrome:
Maximum height/width: 32,767 pixels
Maximum area: 268,435,456 pixels ( 16,384 x 16,384)

Firefox:
Maximum height/width: 32,767 pixels
Maximum area: 472,907,776 pixels ( 22,528 x 20,992)


lpatiny:
There should be a simple mode in which the user select a size (default to A5) and a resolution (default to 300 DPI). From those values we can calculate the number of HORIZONTAL pixels.

The horizontal aspect should no change but we could change the vertical



image

lpatiny:
Save in cm and reload is a mess

image

Do not display decimals for px and it should be written @300 and not @cm dpi

image

lpatiny:
The functionality looks ok to me but for now just hide the toggle Basic / Advanced. We will keep only Basic and wait if somebody really needs the advanced mode.

image

PR #3270, opened 2024-10-16
feat: control the margin top of the peaks label

hamed-musallam:


src/component/reducer/preferences/panelsPreferencesDefaultValues.ts
```diff
@@ -140,11 +140,11 @@ const getPeaksDefaultValues = (
     intensity: { show: true, format: '0.00' },
     fwhm: { show: true, format: '0.00000' },
     mu: { show: false, format: '0.00000' },
+    gamma: { show: false, format: '0.000' },
```
targos, line 143: How do you decide it has 3 decimals?

targos: Ok, then @jobo322 how did you decide? I'm asking since other defaults are either 2 or 5. This is the first occurrence of 3.

hamed-musallam: As I understand it, the format depends on the default values used in the peak shape, so I assume they used this format based on that.

targos:
Do you need feedback from Luc or we can merge this? I'd like to finish https://github.com/cheminfo/nmrium/pull/3271 this morning

hamed-musallam:
I'm waiting for Luc's feedback before merging (Which I prefer), but if you're in a hurry, we can go ahead and merge it now and make improvements later.

lpatiny (approved):

Issue #3233, opened 2024-09-12
Copy raw values


lpatiny:
In a similar manner to copying the table results, we should be able to copy all the meta information, as well as the selected X and Y variables. The columns of the TSV should still be labeled with the variable names.

If could be a 'copy' column with 2 options:
- Copy tab-delimited
- Copy chart

<img width="476" alt="image" src="https://github.com/user-attachments/assets/ab1e8b9c-2f36-468e-8f97-2312f9ab85c4">

lpatiny:
Here is an example of the format:

```
Spectrum Name	Solvent	Pulse	Experiment	x	y
XTC 2	D2O	zg30	1d	-20.420917611073236	-6.87771860712237
XTC 7	D2O	zg30	1d	-7.301988258900954	18.040107806638193
XTC 6	D2O	zg30	1d	13.621030797657086	-68.52101948718956
XTC 4	D2O	zg30	1d	20.791754192347174	11.024381003112337
XTC 8	D2O	zg30	1d	50.28103291095254	-4.965026788677216
XTC 1	D2O	zg30	1d	73.45521103869329	8.990949521465897
XTC 10	D2O	zg30	1d	158.18065123241678	-45.511889584269994
XTC 11	D2O	zg30	1d	698.7369857677671	543.1436330423102
XTC 9	D2O	zg30	1d	897.7837708961251	648.5633659346997
XTC 5	D2O	zg30	1d	9643.43957442702	9665.639130947864
XTC 3	D2O	zg30	1d	10228.690841516114	10236.12341430302

```

Issue #3232, opened 2024-09-12
Order by in multiple spectra analysis


lpatiny:
In multi spectra analysis we have this nice feature that allows to display charts.

One important chart would be the evolution of the purity depending the acquisition time. However today there is no simple way to sort the spectra based on acquisition time (or any other parameters).

<img width="558" alt="image" src="https://github.com/user-attachments/assets/48721377-fb93-4f59-8469-0ceaf61ffaf9">

The idea is that next to the variables X and Y we have a little sort button that allows to sort the spectra based on this specific variable.

PR #3272, opened 2024-10-17
feat: sort spectra by specific parameter

hamed-musallam:
The spectra are sorting without mutating the original state

- https://github.com/cheminfo/nmrium/issues/3232
- https://github.com/cheminfo/nmrium/issues/3233


lpatiny:
Exporting the raw data is working perfectly.

Concerning sorting, is there any reason I can not sort by an integration value ?

image

hamed-musallam:
> Exporting the raw data is working perfectly.
> 
> Concerning sorting, is there any reason I can not sort by an integration value ?
> 
> image

The way we sort the integration is different from how we sort the spectrum parameter.

lpatiny:
Would it be possible to toggle between 3 states ? Sort up / Sort down / reset to original order.

Also would it be difficult to also add the sort by integral ?

lpatiny (approved):
Functionality LGTM

PR #3273, opened 2024-10-17
chore: add 'npm run dupe' to find duplicate packages

lpatiny:


targos:
Unfortunately, it seems to add a gigantic number of new dev dependencies :(

targos:
I could make a small lib with zero deps to do the same thing...

lpatiny:
Indeed they depend on `npm`. Didn't know you could install npm like that in fact.

https://github.com/vovacodes/find-duplicate-dependencies/blob/master/package.json#L29

Issue #3281, opened 2024-10-30
Use SelectedTotal component


stropitek:
Use the react-science component `SelectedTotal` everywhere it can be used.

wadjih-bencheikh18:
This component is only used in PanelHeader. It's better to directly use react-science PanelHeader component.

hamed-musallam:
resolved by [#3283](https://github.com/cheminfo/nmrium/pull/3283)

PR #3283, opened 2024-11-01
Use react-science PanelHeader

wadjih-bencheikh18:
closes : https://github.com/cheminfo/nmrium/issues/3281


wadjih-bencheikh18:
@stropitek while fixing tests. Found that many webkit tests (which are not required) in the main branch are failing should I open an issue for that?

test-e2e/panels/structures.test.ts
```diff
@@ -465,9 +464,16 @@ test('molecules 1H spectrum', async ({ page, browserName }) => {
...
-    ).toBeVisible();
+    if (browserName === 'chromium') {
+      await expect(
+        nmrium.page.locator('_react=MoleculePanel >> text=1 / 3'),
+      ).toBeVisible();
+    } else {
+      await expect(
+        nmrium.page.locator('_react=MoleculePanel >> text=1 / 2'),
+      ).toBeVisible();
+    }
```
stropitek, line 475: Why is there a different result on different browsers?

src/component/panels/MoleculesPanel/MoleculePanelHeader.tsx
```diff
@@ -290,14 +292,6 @@ export default function MoleculePanelHeader({
       </Toolbar>
 
       <div style={{ flex: 1 }}>{children}</div>
-      {molecules && molecules.length > 0 && (
-        <p style={styles.counter}>
```
stropitek, line 294: You are not using this style anymore but I don't think you have removed it from `styles`.

stropitek (commented):
I let @lpatiny review the functionality.

test-e2e/panels/structures.test.ts
wadjih-bencheikh18: It was like that before I just improved the test. 
Maybe @hamed-musallam knows why it's diffrent.

hamed-musallam: At that time, the copy/paste molecule structure function worked well only in Chrome, but this is no longer true.

This feature needs to be tested again, and then these lines should be refactored from the test code

wadjih-bencheikh18: Still working only on Chrome. As you can see in test only chrome have 3 molecules.

stropitek: ok let's leave it like that

hamed-musallam (approved):

PR #3296, opened 2024-11-23
pass logger to filter for time consuming information

jobo322:


src/component/EventsTrackers/KeysListenerTracker.tsx
```diff
@@ -176,6 +178,7 @@ function KeysListenerTracker(props: KeysListenerTrackerProps) {
                     type: 'DELETE_EXCLUSION_ZONE',
                     payload: {
                       zone,
+                      logger,
                     },
```
hamed-musallam, line 182: We need to ensure the reducer function remains pure. Passing the logger to the reducer introduces a side effect, which we should avoid. This issue has been consistently repeated across all the PRs you submitted

hamed-musallam: we should lift the function up that requires the logger as a parameter to the callback function and pass its result down to the reducer

lpatiny: We have a reducer that for example apply apodization on a spectrum. Apodization being done by nmr-processing.

We would like that nmr-processing can log some information (time spend in each filter) in the `logger`.

Problem is that the logger is changing another part of the state and this is not expected to be done in a reducer.

@targos Do you have some suggestions how to solve this issue ?

targos: Hamed said it all. You need to do the processing outside of the reducer and send the result to the reducer. A reducer is not supposed to do heavy work. Just take some data and update the state with it.

lpatiny:
Will not be implemented for now and will need a new PR.
- https://github.com/cheminfo/nmrium/issues/3319

PR #3299, opened 2024-11-25
feat: 2d apodization

hamed-musallam:


hamed-musallam:
@lpatiny 

Could you provide a description for the Apodization Dimension One and Two filter tools, and what are their shortcuts
https://github.com/cheminfo/nmrium/blob/acf2b3632d995925f31df749df94cff9695dd7ef/src/component/toolbar/ToolBar.tsx#L299-L321

PR #3308, opened 2024-12-03
feat: improve manual range picking and fix simulation of diagonal peaks in COSY prediction

jobo322:
- fix the simulation of the peaks in the diagonal in COSY predicted spectra.
- improve auto-processing 2D raw data, apodization
- improve/fix the multiplicity determination by multiplet-analysis in manual range picking:
- * by phase the selected range data before analyze the multiplet
- * reduce criteria value for the selection of J coupling


jobo322:
@lpatiny should we allow the prediction of COSY with cross-peaks with a path length of 4 bonds as max value?

lpatiny:
> @lpatiny should we allow the prediction of COSY with cross-peaks with a path length of 4 bonds as max value?

No we should not. But we have the couplings coming back from the prediction so we should check from the result if there are some couplings or not. I would put a threshold of 1.5Hz in order to display the cosy coupling or not.

So we should not simply use 2J or 3J anymore but use the result from the prediction.

lpatiny (commented):
The COSY prediction seems indeed much more accurate.

It can be tested from:
https://3307-manual-range-picking-re.nmrium.pages.dev/#/SamplesDashboard/8kbdgnwtkbi/Prediction

We now have:

image

And before it was wrongly:

image

lpatiny:
@jobo322 Is this PR related to issues ?

For COSY I agree it works correctly now.
I forgot about improve manual range picking and how to test it.

PR #3320, opened 2025-01-10
refactor: use styled emotion

hamed-musallam:


src/component/1d-2d/FieldEdition.tsx
```diff
@@ -1,7 +1,6 @@
-/** @jsxImportSource @emotion/react */
 import type { PopoverProps } from '@blueprintjs/core';
-import { Popover } from '@blueprintjs/core';
-import { css } from '@emotion/react';
+import { Popover as BasePopover } from '@blueprintjs/core';
```
targos, line 2: Daniel and I think it's better to leave the original imported component name and give a more specific name to the styled version. Could be as simple as `StyledPopover` if you don't have a better idea.

hamed-musallam: The real challenge for me is always coming up with a name for a component where the only difference is the style. Honestly, it feels like naming a twin sibling but trying to convince everyone they are different! 😂

PR #3342, opened 2025-01-28
feat: Inset tool

hamed-musallam:


lpatiny:
- hide grid
- hide J graph

If we have ranges:
- toggle integral
- toggle peaks label
- toggle top spectrum / top peak
- toggle assignment label
- toggle multiplicity tree

If we have integrals:
- toggle integral

If we have peaks:
- toggle peaks label
- toggle top spectrum / top peak


Rename to "Top of the peak" / "Top of the spectrum"

image

PR #3346, opened 2025-01-30
fix: floating molecule position

hamed-musallam:


lpatiny (commented):
The functionality looks perfect to me.

PR #3395, opened 2025-03-04
refactor: assignments context

hamed-musallam:


lpatiny (approved):
The functionality concerning assignment (not testing Michael Wenk code) is still working as expected.

PR #3408, opened 2025-03-13
feat: improve export publication string options

hamed-musallam:


lpatiny (approved):
Currently it changes nothing but this can be fixed later. So this PR can be merged.

PR #3412, opened 2025-03-14
fix: pass start and end in PPM for 'ANALYZE_SPECTRA' action

hamed-musallam:
Resolve a bug introduced by commit #b164290


lpatiny:
There is a new layout problem in the table

image

PR #3415, opened 2025-03-18
refactor: catch and display export error correctly

### LINKED ISSUE #3394: Change window prevents copy to clipboard
When copying large spectra like 2D to the clipboard it may take 1 or 2s

https://dev.nmrium.org/#/SamplesDashboard/dczjpmbs4a5/Viewcytisine2ddatacytisine2dalljson

If during this time you change the focus and go to the word processor to paste the image NMRium will still say it is copied but there is a catch error in the console

Image

And the image is not in the clipboard.

If there is an error please do not show the green feedback that it is copied but report a pink feedback with the error message.

I don't think we need to check if window.clipboard exists because currently I think it is always available and if not you may as well report the error in the pink feedback to the user.

Another similar error can also appear when OffscreenCanvas is not available.

ReferenceError: Can't find variable: OffscreenCanvas
  at pz (./node_modules/nmrium/src/component/utility/export.ts:220:37)

Please also report the error to the user and do not Report this error.

hamed-musallam:
close  #3394

related to #3407

### DIFF
--- a/src/component/utility/export.ts
+++ b/src/component/utility/export.ts
@@ -335,17 +335,26 @@ async function resolveBlob(b: Blob): Promise<Blob> {
 
 async function writeImageToClipboard(image: Blob, isSafari = false) {
   await navigator.clipboard.write([
-    new ClipboardItem({ [image.type]: isSafari ? resolveBlob(image) : image }),
+    new ClipboardItem({
+      [image.type]: isSafari ? resolveBlob(image) : image,
+    }),
   ]);
 }
 async function copyBlobToClipboard(canvas: OffscreenCanvas) {
+  // Check if the document is focused, If it is not focused, throw an error to inform the user.
+  if (!document.hasFocus()) {
+    throw new Error(
+      'Copy failed because the browser is not focused. Please click the window to focus and try again.',
+    );
+  }
+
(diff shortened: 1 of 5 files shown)


src/component/utility/export.ts
```diff
@@ -335,17 +335,32 @@ async function resolveBlob(b: Blob): Promise<Blob> {
...
+    new ClipboardItem({
+      [image.type]: isSafari ? resolveBlob(image) : image,
+    }),
   ]);
 }
 async function copyBlobToClipboard(canvas: OffscreenCanvas) {
+  // Check if the document is not focused. Focus the window.
+  if (!document.hasFocus()) {
+    window.focus();
+  }
```
targos, line 347: Is there a way to reproduce this? In my tests with Chrome, it doesn't bring the focus back.

hamed-musallam: You can disable this code and try Luc's scenario (https://github.com/cheminfo/nmrium/issues/3394): switch to another application immediately after clicking the "Copy" button on the "Copy to Clipboard" dialog.

hamed-musallam: yes you are right, it should be removed

targos (approved):

PR #3436, opened 2025-03-24
refactor: set fixed size to multiplicity tree edges and header

hamed-musallam:
refactor:  set multiplicity tree edge length to 20px
refactor:  set multiplicity tree header label size to 12px


lpatiny (changes requested):
You you try to give as height 10px rather than 20 ?

Could you also prevent the numbers to scale ?

image

hamed-musallam:
This with 10px 
image
image

lpatiny (approved):

PR #3447, opened 2025-03-27
feat: processed apodization

hamed-musallam:


lpatiny:
Looks good thanks !

By default please have the 'Processed' toggle to Off

image

Testing on https://processing-without-back-in-t.nmrium.pages.dev/#/SamplesDashboard/1o4x1711gyy/Viewdatacytisinedatacytisineprocessed13Cjson

The FID vertical scale is not correct. We see nothing

image

Drawing the line for the function would make sense if we keep the same horizontal axis as when displaying the FID. If this is not easily possible it is better to remove it when displaying the processed spectrum. The green line that we see currently means nothing

image

hamed-musallam:
>Drawing the line for the function would make sense if we keep the same horizontal axis as when displaying the FID. If this is not easily possible it is better to remove it when displaying the processed spectrum. The green line that we see currently means nothing

The issue is that the scales are different on the X and Y axes. We have two options:

1- Display the horizontal axis the same way as when showing the FID, but this will cause issues when zooming in, as it will only be visible at full zoom-out.

2- Keep the X-axis for the apodization line fixed, so it remains unaffected by zooming in and out on the spectrum.

hamed-musallam:
> The FID vertical scale is not correct. We see nothing

The issue is related to updating the domain when switching from processed to unprocessed apodization when moving the edition from zero-filling to apodization or vice versa. Based on the domain rule, only the X-domain should be updated for apodization and zero-filling, we face this issue because we do not store the processed flag inside the filter options. As a result, we cannot determine whether the previous filter was processed or not, and as a result, we update the x and y domain

lpatiny (approved):
Functionality looks good to me

PR #3449, opened 2025-03-28
feat: hide 2D phase correction for non-quadrant spectra

hamed-musallam:


lpatiny:
Sorry I didn't express myself correctly. If we want to stay consistent with 1D, we should hide and not just disable the button.

This is the case for the simulated spectra for example:

https://disable-2d-phase-correction.nmrium.pages.dev/#/SamplesDashboard/u78ehnsonk/Viewtestssimulateddatatestssimulatedd2-changingHeighjson

@hamed-musallam  Could you just hide the button ?

lpatiny:
But after you can merge this PR

PR #3452, opened 2025-04-01
fix: update react-science

stropitek:
Fixes bugs related to embedding NMRium in a project using tailwindcss

### DIFF
--- a/src/component/panels/PanelsBar.tsx
+++ b/src/component/panels/PanelsBar.tsx
@@ -19,7 +19,6 @@ const PanelsBarContainer = styled(ButtonGroup)`
   padding: 4px;
   border-left: 1px solid ${Colors.LIGHT_GRAY4};
   background-color: ${Colors.WHITE};
-  overflow: hidden;
 `;
 
 function useHiddenItemsMenu(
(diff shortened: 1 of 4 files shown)


src/component/panels/PanelsBar.tsx
```diff
@@ -14,12 +14,12 @@ import { useTogglePanel } from './hooks/useTogglePanel.js';
 
 const PanelsBarContainer = styled(ButtonGroup)`
   flex-wrap: nowrap;
-  overflow: hidden;
   height: 100%;
   gap: 4px;
   padding: 4px;
   border-left: 1px solid ${Colors.LIGHT_GRAY4};
   background-color: ${Colors.WHITE};
+  overflow: hidden;
```
stropitek, line 22: This property creates a visual bug.

But removing it make the playwright tests fail, and I'm not sure what to do about it.

✅ Without overflow hidden

CleanShot 2025-04-01 at 18 11 02


🔴 With overflow hidden
CleanShot 2025-04-01 at 18 10 46

lpatiny: No problem to update the test. It checks the width of an integral that can indeed change with any small layout change.

alt text

PR #3465, opened 2025-04-14
fix: generate ranges from signals instead to directly use it as range

jobo322:


src/data/data1d/Spectrum1D/ranges/resurrectSpectrum.ts
```diff
@@ -94,7 +99,10 @@ function resurrectSpectrumFromSignals(
           isFt: true,
           name,
         },
-        ranges: { values: signals, options: { sum: 100 } },
+        ranges: {
+          values: signalsToRanges(signalsJoin(signals)),
+          options: { sum: 100 },
```
lpatiny, line 104: Should probably be the sum of the signals integration or the sum of the atoms if available ?

jobo322:
@lpatiny  could you have a look to the databases and the ranges generated. 

https://data.nmrium.org/maybridgeCT/toc.json
https://data.nmrium.org/iclCT/toc.json

src/data/data1d/Spectrum1D/ranges/resurrectSpectrum.ts
```diff
@@ -81,6 +91,17 @@ function resurrectSpectrumFromSignals(
...
     });
+
+    let totalIntegration = 0;
+    for (const signal of signals) {
+      const { nbAtoms, atoms = [] } = signal;
+      if ('nbAtoms' in signal) {
+        totalIntegration += Number(nbAtoms);
+      } else if ('atoms' in signal) {
+        totalIntegration += atoms.length;
+      }
```
targos, line 102: This is a very weird pattern. I would rather do:

```suggestion
      const { nbAtoms, atoms } = signal;
      if (nbAtoms !== undefined) {
        totalIntegration += Number(nbAtoms);
      } else if (atoms !== undefined) {
        totalIntegration += atoms.length;
      }
```

PR #3469, opened 2025-04-17
feat: horizontal and vertical zoom with CTRL + drag

hamed-musallam:


hamed-musallam:
I think it should be available, similar to how other Zoom functionalities which always available, whatever the selected tool.

hamed-musallam:
> Please also add this new zoom feature in the help
> 
> image

Can you suggest input text? The new functionality works for both horizontal and vertical zoom, since it depends on the mouse drag direction.

lpatiny:
image

H / V  CTRL + Drag

PR #3471, opened 2025-04-22
chore: update dependencies

hamed-musallam:


targos:
I will work on adapting the CanvasEditor tomorrow.

targos:
@lpatiny This is ready for testing on your side. I just need to update the playwright tests

hamed-musallam:
Can we improve the editor UI? I think some icons and colors appear inconsistent
image

targos:
Can you elaborate? Your screenshot just shows the toolbar

hamed-musallam:
The icon sizes are inconsistent, and the delete, undo, and rotate icons do not look as good as they used to. For the color, I think black is better than blue. I believe the old icons are more or less better than what we have currently.

image

targos:
Unfortunately, this is out of our control. The editor UI is managed by the upstream Java project.
We would like to work on a different one where the toolbar is no longer drawn in a canvas but that will take time.

lpatiny (approved):
Seems to work perfectly thanks !

hamed-musallam:
What about https://github.com/epam/ketcher, which is run under Apache-2.0 license.

lpatiny:
openchemlib is not just a sketcher but the full library we use for assignments (highlights + diaIDs), advanced structure search, advanced stereochemistry, etc. So we are very bound to this editor.

PR #3494, opened 2025-05-05
feat: reopen the collapsed panels area when toggling the panel from the panels bar

hamed-musallam:


targos:
Are you waiting for feedback on this? I think it's a good idea.

PR #3504, opened 2025-05-07
fix(jcamp1Dexport): add frequencyOffset and spectralWidth info 

jobo322:


lpatiny (approved):

src/data/PredictionManager.ts
```diff
@@ -261,6 +261,13 @@ function generated1DSpectrum(params: {
     nbPoints,
     lineWidth,
   });
+
+  const first = x[0] ?? 0;
```
lpatiny, line 265: @targos What is the best way to solve the undefined issue in this code ?

lpatiny: The problem with TS. Should we make a check that length is >0 and throw otherwise ? Should we simply do `as number` ?

targos: It depends on the situation:
- If the array may be empty, the current code seems good
- If you expect it to never be empty, either throw or use `as number`

PR #3532, opened 2025-05-20
feat: reorder filters

hamed-musallam:


src/component/elements/Sections.tsx
```diff
@@ -1,3 +1,8 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import {
+  draggable,
+  dropTargetForElements,
+} from '@atlaskit/pragmatic-drag-and-drop/element/adapter';
```
stropitek, line 5: This import will make the build not native ESM compatible.
@targos I don't know if it really matters now that we are not using Next.js anymore.

See https://github.com/zakodium-oss/react-science/issues/899 and the related issues and resolution PR.

targos: This is more important without Next.js

hamed-musallam: We can move to another package that supports ESM, like https://github.com/clauderic/dnd-kit

stropitek: I would not use a different library in react-science and here.

Also dnd-kit has a lot of known performance issues and is currently in a big rewrite that is taking a lot of time to complete. I would not use it right now.

hamed-musallam: I do not think that @atlaskit/pragmatic-drag-and-drop will resolve the issue soon, as a related problem has been open in their repository since [April 7, 2024](https://github.com/atlassian/pragmatic-drag-and-drop/issues/27#issue-2229922572) with no updates or activity

hamed-musallam: Then, I will follow your way, and I hope they will resolve the issue in the coming days.

src/component/elements/pdnd.cts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
stropitek, line 1: Is this necessary? You installed the dependency so it is not extraneous.

stropitek (approved):

src/component/elements/pdnd.cts
```diff
@@ -0,0 +1,10 @@
...
+// https://github.com/atlassian/pragmatic-drag-and-drop/issues/27
+// Given how `@atlaskit/pragmatic-drag-and-drop` publishes ESM in a non-native way,
+// we have to trick TS into using the CJS build so that our build is compatible
+// with native ESM.
+// We cannot use `verbatimModuleSyntax` ts config option with this pattern.
+export {
+  draggable,
+  dropTargetForElements,
+} from '@atlaskit/pragmatic-drag-and-drop/element/adapter';
```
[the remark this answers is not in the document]
hamed-musallam: I disable the rule over the file 'pdnd.cts'. Thank you, Daniel.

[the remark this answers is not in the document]
stropitek: What I meant was that you should configure the `import/no-restricted-imports` rule so that it is not allowed to import directly from the package. Except for the `cts` file which can.

And I think the `import/no-extraneous-dependencies` does not seem necessary (there is no extraneous import).

PR #3539, opened 2025-05-22
fix of phase correction 2D

jobo322:


jobo322:
@hamed-musallam I saw the use of tempData, currently I am using it to fix the initial problem, but I am not sure when this tempData is updated at the moment to apply the phase correction filter. Could you help me check it, please?

jobo322:
I think, the use of increments instead of passing the entire value of phase correction parameters would fix the bug. @hamed-musallam I saw a hook related with it, is it not used currently?

hamed-musallam:
resolved by  #3540

PR #3542, opened 2025-05-23
fix: new approach for determining the default contour levels in 2D NM…

jobo322:
…R spectra


lpatiny (commented):
This approach looks too cpu intensive to me.

We need to discuss about it but if the issue is the noise we could think about having some smoothing option in SanPlot.

lpatiny (approved):
Looks good !

PR #3556, opened 2025-05-30
fix: align 1D spectra to the same top when loading

hamed-musallam:


lpatiny:
The current code does not deal correctly with large negative peaks like in 

image

The height of a spectrum should be the absoltue value of all the Y points. 

you can use on all the RE points: https://github.com/mljs/spectra-processing/blob/main/src/x/xMaxAbsoluteValue.ts

hamed-musallam:
@lpatiny 

Could you test again? I think now it works as expected
image

hamed-musallam:
> @lpatiny
> 
> Could you test again? I think now it works as expected image

@lpatiny The same-top action still uses the old code, but I have implemented a new way to have the same behaviour when loading spectra that no longer relies on D3 functions. The only visual difference between the old and the new way that I rescale the top peaks (negative or positive) to 80% of the viewer.

PR #3569, opened 2025-06-02
feat: remove related insets when a spectrum is deleted

### LINKED ISSUE #3552: Removing spectrum will not remove inset
If you have an inset and remove the original spectrum you get an empty inset that can not be deleted

Image

hamed-musallam:
close #3552


lpatiny (approved):
Works perfectly thanks !

PR #3570, opened 2025-06-02
fix: clone matrix options before dispatch

hamed-musallam:


lpatiny (approved):
Works correctly now thanks !

PR #3571, opened 2025-06-02
fix: unify scroll direction for vertical and horizontal zooming

### LINKED ISSUE #3565: Is shift + scroll wheel inverted
When I zoom in vertically, I need to turn the scroll wheel of my mouse in one direction. However, if I shift and scroll, I should turn the scroll wheel in the opposite direction to zoom horizontally.

This seems weird but if I try on a touch pad both zoom goes in the same direction.

@hamed-musallam Any idea what could be the reason ?

hamed-musallam:
close #3565

### DIFF
--- a/src/component/reducer/actions/ToolsActions.ts
+++ b/src/component/reducer/actions/ToolsActions.ts
@@ -329,7 +329,10 @@ function zoomWithScroll(
     scaleX = get2DXScale(draft);
     scaleY = get2DYScale(draft);
   }
-  const scaleRatio = toScaleRatio({ delta: deltaX, invertScroll });
+  const scaleRatio = toScaleRatio({
+    delta: deltaX,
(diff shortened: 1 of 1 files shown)


lpatiny:
I didn't take a mouse with me but now it is with the trackpad that the inversion occurs.
Are you able to distinguish those 2 events ? Mouse wheel versus 2 fingers on trackpad ?

targos:
We cannot really do that distinction. It's the user who decides in which direction the scroll goes (and in macOS for example there is a different setting for mouse and trackpad)

targos:
Refs: https://github.com/w3c/uievents/issues/57
Only Safari exposes this information.

lpatiny:
It is true that I have the same issue on my mac with the 'natural' scrolling option. @targos should this become a user option in NMRium ?

PR #3590, opened 2025-06-11
feat: improve multiple spectra analysis formula calculation 

hamed-musallam:


src/data/data1d/multipleSpectraAnalysis.ts
```diff
@@ -218,7 +218,12 @@ function calculate(
   data: AnalysisRow,
   formula = '',
 ) {
-  const array = formula.split(/[%()*+/-]/);
+  const cleanFormula = formula
+    .replaceAll(/Math\.\w+\(/g, '') // Remove Math.<function name>(
+    .replaceAll(')', '') // Remove closing parentheses )
+    .replaceAll(/\s+/g, ''); // Remove spaces
+
+  const array = cleanFormula.split(/[%()*+/,-]/);
```
lpatiny, line 226: I don't understand why we care about which variable to add. We can just add all the values of all the variables that are not formula. In the future we could add more objects and allow complex formula and this seems difficult to maintain.

PR #3592, opened 2025-06-16
refactor: rename apodization filter title  to 'Apodization / Window functions'

hamed-musallam:


hamed-musallam:
@lpatiny 

We can not complete this PR until we rename the filter in the `nmr-processing` package, which is archived !!!

Issue #3331, opened 2025-01-16
Numerotation of molecules


lpatiny:
In natural products it is required to specify precise numbering (also with prime).

![Image](https://github.com/user-attachments/assets/fe8d0da8-a801-45c5-887c-77476454db91)

Here is an example of molfile with numbering

```

  ChemDraw01162517442D

  8  8  0  0  0  0  0  0  0  0999 V2000
   -1.0717    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -1.0717   -0.8250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.3572   -1.2375    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.3572   -0.8250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.3572    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.3572    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.3572    1.2375    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0717    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0        0
  2  3  1  0        0
  3  4  1  0        0
  4  5  1  0        0
  5  6  1  0        0
  6  1  1  0        0
  6  7  1  0        0
  5  8  1  0        0
V    1 3
V    6 2
V    7 1
M  END
```

PR #3668, opened 2025-08-29
feat: support custom label edition on atom

tpoisseau:
Closes: https://github.com/cheminfo/nmrium/issues/3331


tpoisseau:
Is it good to you @lpatiny ?

lpatiny:
I'm loosing the fact that an atom is assigned.

Take the example

image

then 'r' + Auto ranges picking.

Before when you create a link the atom becomes yellow. Indicated that it has been assigned.



In this PR we are loosing this important feature

lpatiny (approved):

PR #3684, opened 2025-09-05
chose: fix export menu text and reorder them

lpatiny:


lpatiny:
@hamed-musallam Could you add tool tips in the Save data and Save as menu with the shortcut ? I may add more text in the future.
You can also merge this PR once done.

PR #3687, opened 2025-09-05
refactor: combine panel status menu

### LINKED ISSUE #3685: Combine menu
Open on load only make sense if the panel is active.

Image

I think we should remove this column and have 4 choice in the select drop down menu:
- Hidden
- Available
- Visible = current active
- Active = current active + open on load

hamed-musallam:
We remove 'open on load' column and have 4 choice in the select drop down menu:

- Hidden
- Available
- Visible = current active
- Active = current active + open on load

close #3685


lpatiny:
Currently all are active but they should just be visible (except the first one that should be active).

If you prefer we can keep a closer labelling as before:

- hidden
- available
- active
- open

image

PR #3693, opened 2025-09-11
feat: carbon solvent labelling

jobo322:


lpatiny (changes requested):
If I try on ethyl benzene, the auto range picking yields to an error

image

PR #3727, opened 2025-10-01
refactor: hide title block on insets

### LINKED ISSUE #3725: Do not display title block on insets
Title block should only be displayed in the main screen.

Image

hamed-musallam:
close #3725


lpatiny (approved):
Works as expected

PR #3729, opened 2025-10-01
fix: correctly assign both x and y axes in 2D

hamed-musallam:


hamed-musallam:
We have to address this issue in react-ocl-nmr because when we set the width to OCLnmr, it does not take into account the circle radius, which appears over the atom

PR #3732, opened 2025-10-02
chore: update @zakodium/nmr-types and move them to dev dependencies

stropitek:


stropitek:
It breaks the type in a way that forces me to define default workspace settings for the feature that PR is related to.

targos:
We should probably move the generation of the default settings to the core. It shouldn't be a breaking change to add new fields.

stropitek:
I'm closing this because I realized those types should be in production dependencies.

The NMRium package exports types which depend on the `@zakodium/nmr-types` package.

For example:

CleanShot 2025-10-03 at 10 32 05@2x

`NMRiumData` is exported from the package and it depends on `Spectrum` which itself depends on `Info1D` which comes from nmr-types.

PR #3737, opened 2025-10-07
feat: add yellow highlight to assigned 2D signals

### LINKED ISSUE #3718: Add yellow shadow for assigned signals in 2D
Image

hamed-musallam:
close #3718


lpatiny:
It works as expected, thanks !
Please make the shadow much smaller (max 3px wide) because we may have a lot of protons and it will quickly become very crowded.
Please could you also make it more transparent ? I guess we have have once to have settings for this but it is not urgent.

hamed-musallam:
@lpatiny 

Maybe it would be better to change the color of the gray line instead of using a highlighted rectangle.

PR #3790, opened 2025-10-27
fix: multiple call to `readNMRiumObject` into `NMRiumStateProvider`

tpoisseau:
It was due to `dispatchPreferences` change.
I refactor `PreferencesContext`, the dispatch method is no longer put in state managed by the reducer. Instead, a `useMemo` merge the `preferenceState` with `dispatch` to bind it as `PreferencesProvider` `value`.

* tweak `LoggerProvider`, replace ref invoking `new FifoLogger()` at each render, by a `useState` with init method.


targos:
> tweak LoggerProvider, replace ref invoking new FifoLogger() at each render, by a useState with init method.

Just for my understanding, it this related to the fix or just a "drive-by" improvement?

Issue #3792, opened 2025-10-28
Wrong png and svg extension when export <name.{png | svg}.nmrium>


hamed-musallam:
<img width="436" height="750" alt="Image" src="https://github.com/user-attachments/assets/c5b20280-cfca-47a6-a7c6-23c7e1a9d684" />

<img width="696" height="54" alt="Image" src="https://github.com/user-attachments/assets/1cd234a9-bd72-43c5-a94b-6901a0cf7651" />

PR #3793, opened 2025-10-28
fix: put correct extension for `.tsv`, `.svg` and `.png`

tpoisseau:
Closes: https://github.com/cheminfo/nmrium/issues/3792

### DIFF
--- a/src/component/utility/save_as.ts
+++ b/src/component/utility/save_as.ts
@@ -3,7 +3,7 @@ import fileSaver from 'file-saver';
 
 type Extension = `.${string}`;
 
-export interface SaveAsOptions {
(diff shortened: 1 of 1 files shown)


targos:
Question related to API safety and to avoid future mistakes: what is the advantage of keeping the filename and extension optional (defaulting to `experiment.nmrium`)?

tpoisseau:
Do you prefer to keep the wrapper but use a stricter api (no default value at least on extension).
Or do you prefer to get rid of the wrapper and have no abstraction over `'file-saver'` lib usage ?

targos:
SGTM but I would go further and make `name` mandatory string.

PR #3820, opened 2025-11-12
feat: assign range assignment label from atom custom label

hamed-musallam:


lpatiny (approved):
Works as expected, thanks !

PR #3835, opened 2025-11-20
feat: improving and simplifying the 1D assignment

hamed-musallam:
- improving and simplifying the 1D assignment to reduce ambiguity and ensure more consistent behavior.
- Assignments will now be handled exclusively at the signal level, rather than at the range level.
- Ranges will generally contain a single signal, except in cases where the user manually edits and another signal under the same range.


targos:
@hamed-musallam There's a failing E2E test that's easy to reproduce (just draw where there is no peak):

CleanShot 2025-12-04 at 10 34 46

PR #3851, opened 2025-11-25
fix: clear assignments associated with selected molecule atoms

hamed-musallam:


hamed-musallam:
@lpatiny 

What should happen if I try to assign the same signal to atoms in two different molecules? I believe we should prevent this. Is this right? Because in the current use case, it is allowed.

lpatiny:
Currently if we have 2 molecules and one specific signal is assign to atoms of 2 different molecules if we delete one of the molecules, the assignment is removed from the other molecule as well.
This is an edge case but would be nice to support this situation if not too difficult.

PR #3915, opened 2026-01-20
feat: save to NMRium Archive v1 format

tpoisseau:
* remove save to json / json in zip options

### DIFF
--- a/test-e2e/panels/spectra.test.ts
+++ b/test-e2e/panels/spectra.test.ts
@@ -376,7 +376,7 @@ test('Export sources from 1H spectrum', async ({ page }) => {
     const fields = nmrium.page.locator(
       '_react=SaveAsModal >> input[type="radio"]',
     );
-    await expect(fields).toHaveCount(2);
+    await expect(fields).toHaveCount(3);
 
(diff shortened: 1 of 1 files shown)


tpoisseau:
Draft to disable CI, need changes from core to works.

src/component/1d-2d/components/FloatMoleculeStructures/FloatMolecules.tsx
```diff
@@ -40,7 +40,8 @@ function FloatMoleculesInner(props: FloatMoleculeStructuresProps) {
 
   return visibleMolecules.map(([id, moleculeView], index) => {
     const molecule = molecules.find((m) => m.id === id);
-    assert(molecule !== undefined, 'molecule should be defined');
+    if (!molecule) return null;
```
targos: I disagree with this change. It seems to hide a state inconsistency (view contains a ref to a molecule that isn't in the data anymore)

tpoisseau: Ok, I restored the assert and fix the view in reducer

PR #3927, opened 2026-01-28
feat: improve 2D FT spectrum rendering speed

hamed-musallam:
Implement a new algorithm to reduce 2D FT points to 512 before plotting the contours


hamed-musallam:
> I still feel too slow the contour generation. Looks the matrix used is still bigger than 1K x 1K
> 
> It is a JCAMP file with a 2K × 1K size, and the difference in speed is noticeable. [Cyclosporine_2kx1k.dx.zip](https://github.com/user-attachments/files/24969707/Cyclosporine_2kx1k.dx.zip)

@jobo322 

Hello Alejandro, the reduction works well, but the issue is with the speed, where the reduction function is called each time we zoom or change contour options.
```
Original matrix {y: 1024, x: 2048}
Reduced matrix {y: 1024, x: 1024}
```

PR #3957, opened 2026-02-02
refactor: change header to use zod and new form

Sebastien-Ahkrin:


src/component/modal/setting/tanstack_general_settings/validation/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    .string(message || requiredFieldMessage)
+    .min(1, message || requiredFieldMessage);
+}
+
+export function checkUniqueByKey<T>(
+  data: T[] | undefined,
+  checkKey: keyof T,
+  onError: (message: string, path: Array<string | number>) => void,
+  basePath = '',
+) {
```
tpoisseau: I think it will be simpler to use this helper if it takes directly the context of zod indead an `onError` callback. The purpose of this callback seems only to call `ctx.addIssue` with values passed to `onError`.

4 arguments with optional is a lot, Please prefer a options object.

```ts
interface CheckUniqueByKeyOptions<T> {
  data: T[] | undefined;
  key: keyof T;
  ctx: ZodContext; // I let you find the proper type
  /** @default `''` */
  basePath?: string;
}
```

src/component/modal/setting/tanstack_general_settings/general_settings.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <Dialog
+      isOpen={isOpen}
+      onClose={close}
+      style={{ maxWidth: 1000, width: '50vw', minWidth: 800 }}
+      title="General settings"
+      icon="cog"
+    >
+      <form
```
tpoisseau: You should use `Form` component for proper FormContext with layout attribute.

PR #3962, opened 2026-02-05
refactor: prevent dialog from closing when clicking outside

### LINKED ISSUE #3943: Edit new molecule should have double click
To edit the molecule we need to make a double click

However if there is no molecule we need to make only a single click and double open / close the dialog immediately.

Image

I think we should change it to double click and change the message 

"Double click to edit molecule"

hamed-musallam:
close #3943


lpatiny (approved):
Thanks

PR #3992, opened 2026-02-17
feat: add databases tab

Sebastien-Ahkrin:


src/component/modal/setting/tanstack_general_settings/tabs/database_tab.tsx
```diff
@@ -0,0 +1,313 @@
...
+            return (
+              <form.Field key={index} name={`databases.data[${index}].enabled`}>
+                {(subField) => (
+                  <Checkbox
+                    checked={subField.state.value}
+                    value={String(subField.state.value)}
+                    onChange={(event) => {
+                      subField.handleChange(event.currentTarget.checked);
+                    }}
+                  />
```
tpoisseau: `value={String(subField.state.value)}` is confusing and not needed I think.

`src/component/modal/setting/tanstack_general_settings/ui/table.tsx` (main branch) export `CellCheckbox` styled component. You can use it.

Issue #3989, opened 2026-02-17
Unify table form


tpoisseau:
1. Use the new `TableSection`
2. Should we remove the `#` index column ?

PR #3994, opened 2026-02-18
refactor: `TableSettings` based on `@tanstack/react-table` integration from `react-science`

tpoisseau:
Unify table usage with `TableSettings` and `TableSection`

Closes: https://github.com/cheminfo/nmrium/issues/3989

### DIFF
--- a/src/component/modal/setting/tanstack_general_settings/tabs/database_tab.tsx
+++ b/src/component/modal/setting/tanstack_general_settings/tabs/database_tab.tsx
@@ -12,7 +12,7 @@ import {
   CellActionsButton,
   CellCheckbox,
   CellInput,
-  NewTableSettings,
+  TableSettings,
 } from '../ui/table.tsx';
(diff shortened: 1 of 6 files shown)


targos:
Seeing this in the console when I open the dialog:
CleanShot 2026-02-24 at 11 04 17

targos (approved):
> I implemented tdStyle column metadata with getTdProps (I will implement it directly in react-science if you agree)

SGTM

> I implemented an empty state, I don't know if it should be implemented in react-science also.

Probably makes sense to add in react-science

stropitek (commented):
Code LGTM

The validation of the jpath always fails if you edit it (even if you go back to the original value).

Is it possible to reorder rows (you have set a callback for it)?

The auto-selection of the input made it feel like a excel type of table and made me want to use arrow keys to navigate between cells. I think that would make it quiet practical to edit for example just one column for all the rows. Just wanted to comment about this without you having to act on it.

PR #4004, opened 2026-02-25
feat: supports axis customizations

tpoisseau:


lpatiny:
I don't know what means RTL.

For 1D we should know if we are in the mode isFID or isFT.
- isFID: we can only display either point index or the time in second (the current value)
- isFT: we have 3 possibilities: point index, ppm or hz. 

To convert ppm to Hz you can currently multiply the value in ppm by I think info.baseFrequency

@jobo322  Should we use to convert to Hz info.baseFrequency or info.originFrequency ?

For 2D it is something similar but in both dimension depending if it is before or after FT.   Before FT we have both axes in time like in 1D.

For 2D, the units have to be selected individually because for example the JResolv (after processing (FT in both direction)) experiment would put on the real axis `ppm` as unit by default while on the vertical it should be in Hz. For all other 2D experiments by default after FT we would put `ppm` on both axis.

For the indirect dimension if it does not match a nucleus (your regexp) we would not propose any axis modification.  Unit can become pretty weird like in DOSY experiment

image

tpoisseau:
OK, I moved the discussion in a new issue: https://github.com/cheminfo/nmrium/issues/4040

PR #4005, opened 2026-02-25
fix: resolve bug when using & > *:first-child instead of styled

Sebastien-Ahkrin:


tpoisseau (commented):
You should use `<TableSection>`
It handle the rendering of the title, optional description, actions. And you put your table as children

PR #4008, opened 2026-02-25
feat: support whittaker, cubic, and bernstein baseline algorithms

### LINKED ISSUE #3969: add support for 3 new baseline methods available in baseline filter
Names and options of baselines methods:

name: 'whittaker'
options [XWhittakerBaseline](https://github.com/zakodium/nmrium/blob/e66cb34d09917e88a12148a89951e2705f4321c9/packages/nmr/nmr-processing/src/x/xWhittakerBaseline.ts#L8-L12):

```ts
interface XWhittakerBaseline {
  lambda?: number;
  scale?: number;
  maxIterations?: number;
}
```

name: 'cubic'
options: undefined

name: 'bernstein'
options: [BernsteinBaselineOptions](https://github.com/zakodium/nmrium/blob/e66cb34d09917e88a12148a89951e2705f4321c9/packages/nmr/nmr-processing/src/x/xBernsteinBaseline.ts#L10-L46)

hamed-musallam:
- whittaker options: lambda, scale, maxIterations
- cubic: no options
- bernstein options: maxIterations, tolerance, factorStd, learningRate, minWeight

close  #3969

### DIFF
--- a/src/component/header/SimpleBaseLineCorrectionOptionsPanel.tsx
+++ b/src/component/header/SimpleBaseLineCorrectionOptionsPanel.tsx
@@ -194,7 +194,7 @@ function WhittakerFields({
         debounceTime={250}
       />
       <NumberInputField
-        labelProps={{ title: 'Scale:', style: headerLabelStyle }}
+        labelProps={{ title: 'Learning rate:', style: headerLabelStyle }}
         name="learningRate"
         control={control}
         min={1}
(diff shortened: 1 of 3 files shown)


jobo322:
_**CubicSpline(data, options)**_
Robust iterative cubic spline baseline correction that uses anchor points and iterative clipping to avoid pulling the baseline up through peaks.

_Options:_
zones (array): Array of {from, to} ranges defining baseline regions to prioritize for anchor selection.

**header options:**
numAnchors (number, default: Math.ceil(signal.length / 500), clamped to [8, 512]): Number of uniformly-spaced anchor points to seed the algorithm.
maxIterations (number, default: 10): Maximum number of clipping iterations.
_noiseThreshold_ (number, default: 1.0): Multiplier for noise level to determine clipping threshold. Smaller values are more aggressive.

**Only right panel:**
_tolerance_ (number, default: 1e-6): Convergence tolerance as fraction of noise level.
_noiseLevel_ (number, optional): Override automatic noise estimation with a fixed value.
_noisePercentile_ (number, default: 10): Percentile used for automatic noise estimation from signal differences.

Issue #3932, opened 2026-01-29
Use tanstack form and react-science in general preferences


stropitek:
Work on a dev-only separate button / dialog so that the migration can be done in multiple PRs.

We already have all input fields with tanstack. Question mark for:
- Input with datalist
- Radio button field

PR #4017, opened 2026-02-27
fix: too strict validation

tpoisseau:
```
fix: too strict validation (#4017)

best effort to support invalid workspace

Closes: https://github.com/cheminfo/nmrium/issues/4002

fix: the form submit (#4017)

Use a replace strategy for arrays instead of merge

Refs: https://github.com/cheminfo/nmrium/issues/4006

fix: reset form with encoded workspace (#4017)

fix: handleAddWorkspace by parsing form values (#4017)

Closes: https://github.com/cheminfo/nmrium/issues/4006

refactor: header to use withForm (#4017)

instead passing reset and current values props with broken types 
```

### DIFF
--- a/src/component/modal/setting/tanstack_general_settings/tabs/nuclei_tab.tsx
+++ b/src/component/modal/setting/tanstack_general_settings/tabs/nuclei_tab.tsx
@@ -159,7 +159,6 @@ export const NucleiTab = withForm({
         }),
       ];
     }, [Field, deleteNucleus, insertNucleus]);
-    // const fields = useStore(form.store, (state) => state.values.nuclei);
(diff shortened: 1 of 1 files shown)


targos:
Note that I already made a small review pass and opened a couple of sub-issues on https://github.com/cheminfo/nmrium/issues/3932 before this PR

stropitek (approved):
Not a big deal if not easy to resolve, but the DPI field in export seems independent from the basic / advanced mode, but it is reset when we change the mode.

https://github.com/user-attachments/assets/e42da5ac-1ec4-44da-9f4a-cacebac75fed

Issue #4012, opened 2026-02-26
Text being selected


lpatiny:
New preference settings yields to some text selected.

![Image](https://github.com/user-attachments/assets/ffa8b126-e2ba-4647-9445-3369a3c5d6f4)

PR #4033, opened 2026-03-04
fix(Dialog): text being selected on dialog

Sebastien-Ahkrin:
* set minimum value for svg style to 1
* fix selected text in dialog

Closes: https://github.com/cheminfo/nmrium/issues/4012


tpoisseau (approved):
LGTM

Issue #4045, opened 2026-03-09
Rework / Refactor workspace picker and saving to ensure type safety.


tpoisseau:
Seems some tools don't convert internal representation to workspace type.

* Copy should use `workspaceValidation.safeParse` (log error if parsing fail)
* Paste should use `workspaceValidation.safeEncode` (log error if encoding fail)

The structure of components is hard to follow (with lot of deep nesting). It should be flattened and balanced.

`setWorkspaceSetting` callback is delcared with `inputWorkspace: any` argument.

Avoid type casting `as`

PR #4047, opened 2026-03-10
refactor: workspaces picker / actions

tpoisseau:
Since the copy-paste workspace is experimental, I simplified it.
Workspace form values is copied as `WorkspaceWithSource` (JSON string encoding), it use `formValueToWorkspace`. If form values are invalid, toast an error.
 Paste workspace expect a JSON string of `WorkspaceWithSource`. It use `workspaceValidation.encode` to transform the values in clipboard then apply it to the form with `reset`. If there is an error during the clipboard transformation, toast an error.
 
Simplification and safety of workspace copy-pasting allows to copy a workspace into another one. Apply and save a readonly workspace still force the creation of a new workspace.

Closes: https://github.com/cheminfo/nmrium/issues/4045

fix: form reset is properly applied to checkboxes

### DIFF
--- a/src/component/modal/setting/tanstack_general_settings/workspaces/actions.tsx
+++ b/src/component/modal/setting/tanstack_general_settings/workspaces/actions.tsx
@@ -89,7 +89,10 @@ function CopyPasteButtons(props: WorkspacesProps) {
         }),
       )
       .catch((error: unknown) =>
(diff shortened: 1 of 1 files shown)


tpoisseau:
This should be the last merge before https://github.com/cheminfo/nmrium/issues/4032

So if you still found issues in the form (even unrelated to this PR please give feedback / open new issue)

PR #4091, opened 2026-04-15
chore: update ESLint config

targos:


src/component/loader/useLoadFiles.ts
```diff
@@ -83,6 +83,9 @@ export function useLoadFiles(onOpenMetaInformation?: (file: File) => void) {
     [core, dispatchPayload],
   );
 
+  // TODO: Check what's happening here, because this should be the same as
+  //  workspacePreferences since the access is not asynchronous.
+  // eslint-disable-next-line unicorn/consistent-destructuring
   const currentPreferences = preferences.current;
```
tpoisseau: remove currentPreferences and use workspacePreferences should be OK.

src/data/data2d/Spectrum2D/contours.ts
```diff
@@ -111,6 +111,8 @@ function prepareWheel(value: number, options: WheelOptions) {
   const sign = Math.sign(value);
   const direction = invertScroll ? -sign : sign;
 
+  // TODO: This function is doing the bad practice of mutating its parameters.
```
tpoisseau: You can remove the `TODO: `.
I don't think we will change the code, if you think we should, we should open a new issue

package.json
```diff
@@ -26,7 +26,7 @@
     "clean-lib": "rimraf lib",
     "cspell": "cspell lint \"**/*.tsx\"",
     "dev": "vite --host localhost --port 3000 --open",
-    "eslint": "eslint . --cache --max-warnings=98",
+    "eslint": "eslint . --cache --max-warnings=100",
```
tpoisseau: and revert this, related to previous comment.

tpoisseau (approved):
lgtm

PR #4094, opened 2026-04-16
fix: wrong type stored when changing equivalence value in summary panel

michaelwenk:
In summary panel, the type of the entered value from equivalence field cell is stored as a string value which is not correct and causes problems within the calculation. It should be a numeric value.


targos:
The same fix was made independently in https://github.com/cheminfo/nmrium/pull/4096

PR #4095, opened 2026-04-17
fix: ensure active spectrum indexes match internal state indexes

tpoisseau:
Spectra loading is asyncronous which can change the order of spectra in the internal state. This have unexpected behavior if you display 2d spectra but you get the active spectrum by index and it returns a 1d spectrum.


lpatiny (approved):
The code is now working correctly in my tests

targos (approved):

PR #4096, opened 2026-04-20
fix(ranges): do not edit assignment without signal

### LINKED ISSUE #4061: Trying to assign a range without a signal throws an error
From here: https://github.com/cheminfo/nmrium/blob/fe2f0321256feba5eb04163215d039a5751f0ad0/src/component/panels/RangesPanel/TableColumns/SignalAssignmentColumn.tsx#L25

This whole table is really not type-safe.

targos:
Improve type safety around ranges table and handle absence of signal. Also fixes inconsistencies in the edition of numbers (delta ppm).

Closes: https://github.com/cheminfo/nmrium/issues/4061

### DIFF
--- a/src/component/reducer/actions/RangesActions.ts
+++ b/src/component/reducer/actions/RangesActions.ts
@@ -315,7 +315,7 @@ function handleChangeRangeSignalKind(
 
   const rangeIndex = getRangeIndex(spectrum, range.id);
   const _range = spectrum.ranges.values[rangeIndex];
-  if (_range?.signals) {
+  if (_range.signals && range.tableMetaInfo.signalIndex) {
     _range.signals[range.tableMetaInfo.signalIndex].kind = kind;
     updateRangesRelativeValues(spectrum);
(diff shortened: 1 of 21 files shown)


src/component/reducer/actions/RangesActions.ts
```diff
@@ -315,7 +315,7 @@ function handleChangeRangeSignalKind(
 
   const rangeIndex = getRangeIndex(spectrum, range.id);
   const _range = spectrum.ranges.values[rangeIndex];
-  if (_range?.signals) {
+  if (_range.signals && range.tableMetaInfo.signalIndex) {
```
stropitek: Can `signalIndex` be 0? Then the check might be wrong.

stropitek (approved):

src/component/reducer/actions/RangesActions.ts
targos: probably, good catch!

PR #4114, opened 2026-04-30
feat: Improve peak shape options

### LINKED ISSUE #4092: Determine peak width
@jobo322 could you check why this can not be optimized further ?

Image

hamed-musallam:


src/component/panels/PeaksPanel/PeaksPanel.tsx
```diff
@@ -146,29 +163,6 @@ function PeaksPanelInner(props: PeaksPanelInnerProps) {
...
-        onClick: () => toggleViewProperty('showPeaksSum'),
-        active: showPeaksSum,
-      },
-      {
-        icon: <FaThinkPeaks />,
-        tooltip: 'Optimize peaks',
-        onClick: optimizePeaksHandler,
-      },
-    );
-  }
```
tpoisseau, line 171: I did not found the related issue, So I can't validate removing the experimental flag check without more information.

src/component/reducer/preferences/panelsPreferencesDefaultValues.ts
```diff
@@ -160,6 +161,11 @@ const getPeaksDefaultValues = (
     showDeleteAction: true,
     showEditPeakShapeAction: true,
     showKind: true,
+    defaultPeakShape: {
+      kind: 'pseudoVoigt',
+      fwhm: 1,
+      mu: 0.5,
+    },
```
tpoisseau: Maybe the core should export a `defaultPeakShape` const.

tpoisseau (commented):
Otherwise LGTM except this 

https://github.com/user-attachments/assets/6e01c652-5309-499b-9599-7c6e634d6b5c

Is it normal we can optimize more than 9 times with peak intensity change ?

src/component/panels/PeaksPanel/PeaksPanel.tsx
hamed-musallam: it is just remove the experimental flag and nothing more

src/component/reducer/preferences/panelsPreferencesDefaultValues.ts
hamed-musallam: I’m not sure this is the right way. I would prefer to keep it on the NMRium side rather than in the core package, as these are experimental parameters that may change frequently over time.

src/component/panels/PeaksPanel/PeaksPanel.tsx
tpoisseau: @lpatiny @targos is it ok for you ? The state of theses feature peak shape, sum, and optimize is ready to not be hidden behind experimental flag ?

src/component/reducer/preferences/panelsPreferencesDefaultValues.ts
tpoisseau: https://github.com/cheminfo/nmrium/pull/4114#discussion_r3194075008

If we keep the feature behind experimental flag, I'm ok to keep it like that. If we remove the experimental flag check, let's consider the feature is stable so the core should export the default value.

tpoisseau:
```
chore: Publish new release
 - nmr-processing@22.11.0
 - @zakodium/nmrium-core-plugins@0.7.24
 - @zakodium/nmrium-core@0.7.19
 ```
 
You can upgrade theses packages.

src/component/reducer/preferences/panelsPreferencesDefaultValues.ts
tpoisseau: https://github.com/cheminfo/nmrium/issues/4092#issuecomment-4289104221

> @hamed-musallam Remove the experimental flag of this feature.

But we already merge the PR from mirror, so let's keep it like that now. I'll add a new issue.

src/component/panels/PeaksPanel/PeaksPanel.tsx
targos: I am not aware about these features.

PR #4133, opened 2026-05-15
feat: Improve peak optimization

jobo322:
Implement the optimization on demand to improve speed and optimized shape parameters.

10 peaks takes around 2.8 seconds with an acceptable fit:

improveOptimizationResult


jobo322:
This PR need a new version of nmr-processing with the last version of ml-gsd

PR #4134, opened 2026-05-18
feat: copy peaks table as TSV to clipboard

### LINKED ISSUE #4131: Add export peaks feature
In the peak tab we would need a new copy to clipboard feature:

Image

hamed-musallam:


lpatiny (approved):
Works as expected, thanks !

PR #4151, opened 2026-06-02
compute the from-to range value based in shape factor 

jobo322:


jobo322:
it looks strange to have commits from @hamed-musallam in this branch, so I will open a new PR in the correct checked out

Issue #4162, opened 2026-06-11
Migrate forms off of react-hook-form/yup


targos:
Migrate to `useForm` (react-science) and `zod`

PR #4169, opened 2026-06-16
refactor: use tanstack & zod on `Save As`

Sebastien-Ahkrin:
Refs: https://github.com/cheminfo/nmrium/issues/4162


tpoisseau (approved):
LGTM.
It would be great if you can add autoFocus on name field.

I'm pleased ENTER submit the form (when it is focus), and escape close / cancel the form.

Here is a suggestion if you continue on this issue:
List forms in the application and create one sub-issue by form ;-)

PR #4214, opened 2026-06-19
feat: processings panel for experimental spectra operations

tpoisseau:


targos (approved):

tpoisseau:
For some reasons (I don't know why...) peak picking generate less peaks than before.

I don't know what to do.

On main branch: 
CleanShot 2026-07-01 at 11 38 11@2x

On this branch:
CleanShot 2026-07-01 at 11 38 34@2x

I consider peaks much more pertinent in this branch than before.
Reminder: I did not touch anything related to peaks on this branch. but core had been updated.

I suggest reducing the expected number of peaks to 23 (the number of peaks currently generated with this setup) in the test file.
Do you agree with this plan @targos ?

targos:
This includes an update of nmr-processing, which contains https://github.com/zakodium/nmrium/pull/807
This seems like an expected improvement from that PR.

PR #4232, opened 2026-06-29
feat: support custom tick and axis label styles based on user preferences

### LINKED ISSUE #4225: need to customize the axis label
Currently it is possible to customize the axis ticks

Image

However there is no way to define the axis label. We could either duplicate the current settings for the axis label or use the axis ticks as the format for axis label.

Image

hamed-musallam:


lpatiny:
Nice improvements, thanks !

I was still a little bit confused by the fact that the color of the axis is in fact the color of the Primary ticks and I didn't guess.

I would suggest to have one more section with Axis and the corresponding

image

Just over the section with the Primary ticks

image

lpatiny (approved):
Functionality works perfectly, thanks !

tpoisseau:
- @zakodium/nmrium-core-plugins@0.7.53
 - @zakodium/nmrium-core@0.7.40

PR #4237, opened 2026-06-30
fix: do not look for matching signals if the spectrum does not exist

### LINKED ISSUE #4210: Summary panel filter crash the system

hamed-musallam:


src/component/panels/SummaryPanel/SummaryPanel.tsx
```diff
@@ -153,6 +153,9 @@ function SummaryPanel() {
               link.experimentID,
               true,
             ) as Spectrum2D;
```
targos: If it's expected that the spectrum can be missing, remove the type cast here and in the other places.

PR #4241, opened 2026-06-30
feat: remove peak limit for optimization and adjust autoPeakPicking parameters

jobo322:


lpatiny:
[14.zip](https://github.com/user-attachments/files/29607003/14.zip)

Peak picking on this spectrum does not work at all.

image

Take care also that if we select the 50 more intense peak it should be the more intense in absolute value if both top / bottom are selected.

lpatiny:
I think the issue is that there is still a differentiation between positive max and negative max. The max should be the maxAbsValue.

PR #4242, opened 2026-07-01
feat: manually add signal in a range from 2D

### LINKED ISSUE #4206: Manually add signal in a range from 2D
Add signal from 2D. 

In 2D we often can distinguish between chemical shifts of overlapping peaks. If the range has been created in the 1D we should be able to add signals from the trace in 2D by using maybe a simple click. This will therefore add a new grey line that can be assigned.

hamed-musallam:


lpatiny (approved):
Great and super useful feature ! I really like it !!!

The functionality looks perfect to me. Maybe just remove the value of the integral 

Image

PR #4254, opened 2026-07-08
feat: apply processings from panel change

tpoisseau:
fix: spectrum init when it imply `processings`

* now it check if originalInfo / originalData exists. Only set them if not exists
* disable filters apply if `processings`


src/component/panels/hooks/use_processings_mutation.ts
```diff
@@ -0,0 +1,142 @@
...
+      spectrum.processings = spectrum.processings.map((p) => ({ ...p }));
+    }
+
+    return { spectrum, indexSpectrum };
+  }
+
+  async function submit(spectrum: Spectrum, index: number) {
+    setIsLoading(true);
+    const processedSpectrum = await core
+      .processSpectrum(spectrum)
```
targos, line 58: This is going to change the deep properties of `spectrum`, which are part of the current state, so this is a state mutation, something that we should avoid.
I'm even a bit surprised that it works. I thought `immer` was deep freezing the objects to avoid making mutations by mistake.

targos:
I find a lot of issues by playing with the interface but I'm not sure I should report them here.

src/component/panels/hooks/use_processings_mutation.ts
tpoisseau: We are not in an immer context. I Added this API hook because processings is async immer produce does not support async.

targos: I'm not suggesting that we are in immer context. Immer freezes the output of the producer: https://immerjs.github.io/immer/freezing/

You don't answer to the main concern (my first sentence).

tpoisseau: Oh, I ignore the auto-freezing feature of the result.
That could explain why I had to map processings to create new objects so `error` props could be reseted by core `processSpectrum` method.

tpoisseau: I'll dig through `createDraft` / `finalizeDraft` so. but immer warns it is an anti-pattern :/

targos:
onChange should not duplicate the processing:
CleanShot 2026-07-08 at 16 24 13

targos:
Reordering by drag and drop doesn't seem to work.

targos:
After "Delete all", I see a small horizontal line instead of the FID spectrum.

src/component/panels/hooks/use_processings_mutation.ts
targos: You don't have to use immer. The important part is to not mutate the object returned by `useReducer` (the state and its children objects)

tpoisseau: deep slice strategy seems work properly

targos (approved):

tpoisseau:
https://824-implement-processings-ap.nmrium.pages.dev/#/SamplesDashboard/o5c3l3e2gm7/ViewdatacoffeedatacoffeeCoffeejson

CleanShot 2026-07-09 at 11 55 45@2x

https://dev.nmrium.org/#/SamplesDashboard/f209fsf86vo/ViewdatacoffeedatacoffeeCoffeejson

CleanShot 2026-07-09 at 11 57 14@2x

Do you have an idea of what could cause this issue in my branch @targos ?

targos:
What's the issue?

tpoisseau:
Coffee name don't match, CI expect 1, 2, 3, ...

targos:
I don't know, but I guess it's related to the changes in `src/data/data1d/Spectrum1D/initiateDatum1D.ts`

tpoisseau:
`git checkout main -- src/data/data1d/Spectrum1D/initiateDatum1D.ts`

Solves this issue, but broke processings :/

tpoisseau:
coffee file is a very old nmrium file (unversioned), where display holds a name attribute.

> changes from version 3 to version 4:
> 6- move 'name' object from the 'display' object to  the'info' object

Due to nmrium component responsible of spectrum normalization and responsible to run filters. it was not an issue until now.
Core already do lot of normalization since nmr-load-save legacy package. And now the core is responsible to running the processing before the nmrium component receive it.

I'l try to understand how core, with read and migrations should be updated to have originalInfo name identical to info name in this context.

PR #4256, opened 2026-07-08
feat: change cursor based on the selected tool

### LINKED ISSUE #4251: Cursor change depending the tool
There is currently a big confusion about what will happen depending if we press SHIFT or not.

We checked how gimp was working and we find they have a good idea. Next to the pointer they have an icon showing the action that will be done.

It would be nice to have the same feature in NMRium and display always the crosshair as currently but in the bottom right corner the currently selected action.

Like that if we are in the range analysis and we press shift the icon will change from

Image

to

Image

hamed-musallam:


targos:
When I have the peak picking tool selected and press cmd key on mac, it changes to the zoom icon but it won't zoom (zoom only happens with shift key)

hamed-musallam:
> When I have the peak picking tool selected and press cmd key on mac, it changes to the zoom icon but it won't zoom (zoom only happens with shift key)

Thanks for your feedback! I have fixed the issue.

lpatiny (approved):
For me this is a huge improvement, thanks !

Maybe we will find some edge cases but the current functionality is really great !

PR #4265, opened 2026-07-13
feat: improve signal add/move visual feedback

### LINKED ISSUE #4258: In 2D change cross pointer 1D / 2D
On the top just vertical line
On the left just horizontal line
On the 2D both lines

### LINKED ISSUE #4259: Improve signal addition in 1D / 2D
To better distinguish between new signal and moving an existing one, new signals will have a green dot. Moving signals keep their green color.

If we are in the mode to add / move a signal the normal crosshair disappear and if we move an existing one we have the green dotted line, if it is adding a new signal it is also green dotted line and a green dot.

hamed-musallam:


lpatiny:
Looks good !

when adding a new signal could you also have the green dotted vertical line (like when moving an existing signal). Otherwise it looks perfect to me, thanks.

image

PR #4273, opened 2026-07-15
fix: improve 2D signal guide line behavior

### LINKED ISSUE #4267: Improve add signals for 2D improvements
In 2D could you prolongate the vertical green dotted line when moving or adding new signals ?

When moving the existing signal on the top spectrum the crosshair should disappear

Image

for the vertical spectrum we should also in both case (add / move) have only the green dotted line that goes through the full spectrum.

hamed-musallam:


lpatiny (approved):
Perfect, thanks !

PR #4275, opened 2026-07-16
fix: move signals into their new ranges when cutting a range

### LINKED ISSUE #4276: Cut range does not move signals to new ranges

hamed-musallam:
- https://github.com/cheminfo/nmrium/issues/4276


lpatiny:
I think you forgot the case in which there is no signal in the new range. It should add one in the middle seems to me. Currently we can generate a range without signal.

image

lpatiny (approved):
Works as expected, thanks.

PR #4281, opened 2026-07-21
feat: show processings in error

tpoisseau:
* Operation item (not the expanded part) in error have a red background
  * green when is open (not in error)
  * greyed-out when is after the opened one
  It is a restoration of the filters controls
* display the error message at the end of the Expanded section
* fix live procesings reset. Handle external close with effect to reset  live processings.

### DIFF
--- a/src/component/panels/filtersPanel/processings/processing_item.tsx
+++ b/src/component/panels/filtersPanel/processings/processing_item.tsx
@@ -85,14 +85,18 @@ export function ProcessingItem(props: ProcessingItemProps) {
     liveEditRef.current = liveEdit;
   }, [liveEdit]);
 
+  useEffect(() => {
+    if (!isOpen) return;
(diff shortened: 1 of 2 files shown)


src/component/panels/filtersPanel/processings/processing_item.tsx
```diff
@@ -123,6 +124,14 @@ export function ProcessingItem(props: ProcessingItemProps) {
           operation={operation}
         />
       }
+      headerStyle={getHeaderStyle({
+        isOpen,
+        operation:
+          isOpen && liveEdit.value?.checked
+            ? liveOperationRef.current
```
targos: You should not use a ref during render as it's not reactive.
Please take note of this in an issue if you don't want to fix now.

targos (approved):

PR #4287, opened 2026-07-24
feat: supports `ProcessingOperatorUI.Tool`

tpoisseau:
Experimental: true


src/component/context/processings_mutations_context.api.ts
```diff
@@ -83,9 +94,25 @@ export function useProcessingsMutations() {
...
+      type: 'SELECT_PROCESSING_OPERATOR',
+      payload: { operatorId: undefined },
+    });
+  }
+
   // --- API --- //
 
-  async function apply(
+  const resetLiveChange = useEventCallback(function resetLiveChange() {
+    dispatch({
```
targos, line 107: This only works because we know that `dispatch` will never change? I'm not a big fan of using an API that lets us easily close on outdated deps by mistake.

tpoisseau: No, it works because `useEventCallback` return a memoized callback. Under the hood it use a ref to store our callback, and update it with a useEffect.

https://usehooks-ts.com/react-hook/use-event-callback

targos (approved):

PR #4293, opened 2026-07-31
fix: avoid structuredClone on spectrum data

tpoisseau:
It can crash on big datasets (>500Mo)


tpoisseau:
Loading `kDrive/Projects/NMRium/test_data/huge/HSQC_16k_4k.zip` (720Mo) works in chrome now (even if it is really slow 😅)

targos (approved):

PR #4294, opened 2026-08-03
refactor: assign and highlight the selected signal in 1D/2D traces

### LINKED ISSUE #4269: How to highlight the signal and not always the full range
Image

Currently we can see if we are assigning one signal or the full range.

hamed-musallam:


lpatiny:
Works perfectly, thanks !

If possible when hover an atom it would be nice to highlight using the red dash line the exact position of the signal but the current functionality is good enough.

image

PR #4317, opened 2026-08-19
feat: improve min level determination contours

jobo322:
The idea is to take care of T1 noise ridges, like:

image

this T1-noise does not follow the gaussian distribution of white noise and appear over the noise level.

uses the acquisitionScheme to determine if magnitudeMode should be used for noise level determination when a processed spectra is loaded.


lpatiny (approved):
The functionality looks better, thanks.

PR #4321, opened 2026-08-20
refactor: Table component to TanStack v9

### LINKED ISSUE #1889: Move to react-table version 9
@lpatiny

After react-table version 7, they completely change how it works, They rewrite the library and improve dramatically paunch of features and most important the performance became better.

but we still do not have a clear guide for the migration.

### LINKED ISSUE #4332: Ordering of spectra table
Spectra table can't be ordered by clicking on column header anymore.

targos:
Closes: https://github.com/cheminfo/nmrium/issues/1889


targos:
@hamed-musallam I ask you for review since the modified code was mostly written by you. I tried to keep the API as close as possible to the previous one to avoid having to do too many changes on the tables.

src/component/panels/IntegralsPanel/IntegralTable.tsx
```diff
@@ -133,55 +132,42 @@ function IntegralTable(props: IntegralTableProps) {
...
-      const { showWhen, ...colParams } = col;
-      if (dlv(integralsPreferences, showWhen)) {
-        addCustomColumn(columns, colParams);
-      }
+      );
     }
-
-    columns.sort((object1, object2) => object1.index - object2.index);
     return columns;
-  }, [COLUMNS, integralsPreferences]);
```
hamed-musallam, line 184: I think the original pattern using dlv in one place to filter the columns is cleaner than repeatedly checking dlv(...) while building the columns.

hamed-musallam (commented):
I think the original pattern, where we use `dlv` in one place to filter the columns, is cleaner than repeatedly checking `dlv(...)` while building the columns. 

It could also be turned into a general function for filtering table columns based on user preferences.

hamed-musallam:
Otherwise, the functionality LGTM. Thank you.

targos:
> I think the original pattern, where we use dlv in one place to filter the columns, is cleaner than repeatedly checking dlv(...) while building the columns.

I explicitly added TODOs about using the built-in functionality for column filtering. I didn't want to introduce another workaround in the mean time.

hamed-musallam (approved):

Issue #4221, opened 2026-06-24
Change the export width (advance mode) to nothing, will set NaN to the other linked element


Sebastien-Ahkrin:
<img width="800" height="499" alt="Image" src="https://github.com/user-attachments/assets/6f8c0e66-ab84-4835-9cc2-35fe5066a147" />

hamed-musallam:
Fixed by #4326

PR #4326, opened 2026-08-25
 fix(export): prevent crashes during export size validation

hamed-musallam:
Switching to Advanced mode could crash the app, clearing a size field could break the aspect ratio, and oversized exports failed silently with a cryptic error.

- Fix crash when applying default values in Advanced mode
- Do not reset empty/invalid fields to 1
- Validate against the final rendered pixel size, not the raw input values, and reject sizes the browser cannot handle: less than 1 px per side, more than 32,767 px per side, or more than 268,435,456 px total
- Fix broken size validation logic
- Show clear error messages in the export dialog

### DIFF
--- a/src/component/elements/export/ExportOptionsModal.tsx
+++ b/src/component/elements/export/ExportOptionsModal.tsx
@@ -1,4 +1,5 @@
-import { DialogFooter } from '@blueprintjs/core';
+import { Callout, DialogFooter } from '@blueprintjs/core';
+import { revalidateLogic } from '@tanstack/react-form';
 import { useMemo } from 'react';
 import { AppForm, Button, useForm } from 'react-science/ui';
 import { z } from 'zod/v4';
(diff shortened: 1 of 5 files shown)


hamed-musallam:
- https://github.com/cheminfo/nmrium/issues/4221

hamed-musallam:
@Sebastien-Ahkrin 

Could you review the changes after the refactoring you did for the export modal? I think everything is working well now.

tpoisseau (approved):
I'm not sure to prefer `min={0}` over `min={1}`

Otherwise, LGTM

Sebastien-Ahkrin (approved):
LGTM Thanks you !

hamed-musallam:
> I'm not sure to prefer `min={0}` over `min={1}`
> 
> Otherwise, LGTM

I think a minimum value of 1 is required for values in pixels, but for other units, we could allow values greater than 0.

tpoisseau:
> I think a minimum value of 1 is required for values in pixels, but for other units, we could allow values greater than 0.

Ok, I have no strong opinion on this topic and I approved. Do what you think is better ;-)

PR #4330, opened 2026-08-26
fix: render secondary ticks only between primary ticks

### LINKED ISSUE #4329: Secondary axis ticks render outside primary axis bounds in 2D view
In the 2D spectrum view, secondary ticks are rendered outside the primary axis bounds on edge positions.
This is visible on the top edge and similarly on the horizontal axis edge, where one secondary tick appears outside the expected in-axis direction.

Image

hamed-musallam:
- Remove edge secondary tick rendering that could appear outside axis boundaries.
- Compute secondary tick positions as midpoints between primary ticks.


tpoisseau (approved):
LGTM

PR #4342, opened 2026-08-27
refactor: change useForm on `ImportPublicationStringModal`

Sebastien-Ahkrin:
- **refactor: change useForm on ImportPublicationStringModal**
- **refactor: use form state for logs**
- **refactor: rename variable**


Sebastien-Ahkrin:
Closes: https://github.com/cheminfo/nmrium/issues/4178

src/component/modal/ImportPublicationStringModal.tsx
```diff
@@ -122,21 +142,38 @@ function InnerImportPublicationStringModal(
             info: { nucleus, solvent, frequency, name: parts[0] },
           },
         });
+
         hideLoading();
       });
```
tpoisseau: Can you try to remove the setTimeout ? If it produce unexpected behavior you can continue to it but add a comment to describe why this setTimeout is needed.

src/component/modal/ImportPublicationStringModal.tsx
```diff
@@ -122,21 +142,38 @@ function InnerImportPublicationStringModal(
...
+
       onClose();
-    })();
-  }
+    },
+  });
 
-  function handleOnChange(value: any) {
-    loggerRef.current.clear();
+  const isValid = useSelector(form.store, (store) => store.isValid);
```
tpoisseau: `useSelector` can be quite aggressive on re-rendering. So it should be avoided whenever it's possible.

I suggest you to use `form.Subscribe` component around the Submit button.

Same for `useSelector` for logs (L176).

A single `<form.Subscribe>` should be able to compute the entire logic for `disabled` prop of `form.SubmitButton`.

src/component/modal/ImportPublicationStringModal.tsx
```diff
@@ -122,21 +142,38 @@ function InnerImportPublicationStringModal(
...
+    return () => {
+      loggerInstance.removeEventListener('change', handleLogs);
+    };
+  }, [form]);
+
+  const debounceChanges = useMemo(
+    () =>
+      debounce((value: any) => {
+        resurrect(value, { logger: loggerRef.current });
+      }, 250),
```
tpoisseau:
```suggestion
      debounce((value: string) => {
        resurrect(value, { logger: loggerRef.current });
      }, 250),
```

tpoisseau (commented):
This kind of refactor is also the opportunity to modernize the component and refactor to follow our best practices.

We should use `@emotion/styled` in favor of `style` prop.

Here we use style prop on `StandardDialog`, div first descendant of `StyledDialogBody` and first descendant of this div.

Also I'm pretty sure this jsx tree could be simplified to remove the first descendant of `StyledDialogBody`

tpoisseau:
CleanShot 2026-08-28 at 11 58 08@2x

targos (commented):
Why are the logs part of the form now?

src/component/modal/ImportPublicationStringModal.tsx
```diff
@@ -148,62 +185,65 @@ function InnerImportPublicationStringModal(
...
+                    columns={COLUMNS}
+                    data={field.state.value}
+                    emptyDataRowText="No Logs"
+                    rowStyle={handleRowStyle}
+                    style={{ height: '120px' }}
                   />
-                );
-              }}
-            />
+                </GroupPane>
```
tpoisseau: I consider we have better components About table and table section.
Our reference should be `src/component/modal/setting/tanstack_general_settings` for tanstack form and table usage from react-science.

So I suggest you to use `TableSection` and `TableSettings` (or similar if theses integration of tanstack form and tanstack table are too stricktly tied to the genera settings form).

Sebastien-Ahkrin:
> Why are the logs part of the form now?

like said in direct message on kChat, I choose to use tanstack state to not duplicate source of state (tanstack + useState).

I think, its easier to read and maintain like that

tpoisseau (approved):
LGTM

targos:
The logs accumulate now. Before, they were replaced on change.

Sebastien-Ahkrin:
> The logs accumulate now. Before, they were replaced on change.

I will check that

PR #4343, opened 2026-08-28
fix: prioritize range analysis when Alt is pressed alone

### LINKED ISSUE #4333: Shift in range bug

hamed-musallam:
The issue introduced in https://github.com/cheminfo/nmrium/commit/38819d20d860dc0ce062dc489146498129c545a2


lpatiny (approved):
Bug fixed thanks.

PR #4344, opened 2026-08-28
fix(database): avoid cloning Immer proxies during spectrum resurrection

### LINKED ISSUE #4335: Not able to add spectra from solvent database
Image

hamed-musallam:
The issue introduced in #4254

Database spectrum resurrection copied `info` from an Immer draft. Nested arrays and objects (`first`, `last`, `symbols`, and `varDim`) are draft proxies, which cannot be passed to `structuredClone`.

Create a plain snapshot with `current(activeSpectrum)` before copying its metadata, allowing `initiateDatum1D` to preserve `originalInfo` safely.


lpatiny (approved):
The bug of resurrection of spectra from DB is now fixed. Thanks.

PR #4345, opened 2026-08-28
refactor(auto-peak-picking): change default peak limit to 500

### LINKED ISSUE #4336: Default number of peaks to 500
Image

hamed-musallam:


targos:
Do you have a justification for https://github.com/cheminfo/nmrium/pull/4345/changes/9670760b1c88ce261b435e970798021ed2345a1c

Last time I checked, this error was related to connectivity issues between GitHub and the prediction service. Increasing the timeout would not fix that and only make the run longer when there are issues.

hamed-musallam:
It worked fine locally, but not in the GitHub workflow. I noticed only the workflow test was taking longer, so I increased the timeout and nothing else.

Now that you say it is not the reason, I will try again and revert my changes if that is not the case.

targos:
The failing workflows always upload a Playwright report for local analysis. They are usually detailed enough to know what happens.

PR #4350, opened 2026-09-01
fix(processings): avoid to update view during live change

tpoisseau:
Experimental: true


tpoisseau:
I try to make the CI pass, but it seems github runners have issues to call the prediction service :
CleanShot 2026-09-01 at 11 56 37@2x

PR #4351, opened 2026-09-01
test: mark `automatic assignment panel` flaky

tpoisseau:


test-e2e/flaky.ts
```diff
@@ -0,0 +1,19 @@
...
+  process.env.CI !== '0';
+
+/**
+ * Mark a test as flaky
+ * Inspired by https://github.com/microsoft/playwright/issues/27653#issuecomment-3775359315
+ * @param info
+ * @param reason
+ */
+export function flaky(reason: string, info: TestInfo) {
+  info.skip(
```
targos, line 15: So, skip somehow throws an error to prevent the test from running? Or it will still run be be marked as skipped after?

tpoisseau: It is the same implementation than for nmrium-app.

I don't know how playwright handle skip on runtime but it support it and document it.
https://playwright.dev/docs/api/class-test#test-skip

Community guidelines seems to be using the skip at the beginning of the test, before navigation or assertions.

https://runebook.dev/en/docs/playwright/api/class-testinfo/test-info-skip-2
https://testautomationmastery.com/mastering-conditional-test-skipping-in-playwright/

targos (approved):

PR #4378, opened 2026-09-16
chore: update ml-peak-shape-generator to v5.6.0

jobo322:


src/component/elements/Sections.tsx
```diff
@@ -1,7 +1,7 @@
+import type { Edge } from '@atlaskit/pragmatic-drag-and-drop-hitbox/types';
```
targos: Why?

hamed-musallam: TS shows a deprecation warning, and it says we should import it directly from @atlaskit/pragmatic-drag-and-drop-hitbox/types instead.

targos: Then it should be updated in https://github.com/zakodium/pdnd-esm/blob/main/src/index.cts

hamed-musallam: Can you give the right privileges to push a PR in the pdnd-esm repo ?

targos: Done

hamed-musallam: Can you review https://github.com/zakodium/pdnd-esm/pull/10

